Issue #4, opened 2016-08-19
API: Provide public debugPrint(Ln)


FokkeZB:
All [examples](https://github.com/TheThingsNetwork/arduino-library/tree/master/examples) define:

```c
#define modemSerial Serial1
#define debugSerial Serial
// ..
#define debugPrintLn(...) { if (debugSerial) debugSerial.println(__VA_ARGS__); }
#define debugPrint(...) { if (debugSerial) debugSerial.print(__VA_ARGS__); }
// ..
ttn.init(loraSerial, debugSerial);
// ..
debugPrintLn("Setup for The Things Network complete");
```

The library also [defines](https://github.com/TheThingsNetwork/arduino-library/blob/master/src/TheThingsNetwork.cpp#L7) these functions and uses the `debugSerial` passed to `init()`.

Could we make the debugPrint(Ln) of the library public so that sketches simplify to:

```c
#define modemSerial Serial1
#define debugSerial Serial
// ..
ttn.init(loraSerial, debugSerial);
// ..
ttn.debugPrintLn("Setup for The Things Network complete");
```

FokkeZB:
For the same reasons we didn't do #5 we shouldn't do this either.

PR #12, opened 2016-09-02
Compiler directive for debug output

Nicolasdejean:
Adding debugPrintln and debugPrint to TheThingsNetwork.h to simplify example codes. #4

### DIFF
--- a/src/TheThingsNetwork.h
+++ b/src/TheThingsNetwork.h
@@ -23,6 +23,19 @@
 #define HEX_CHAR_TO_NIBBLE(c) ((c >= 'A') ? (c - 'A' + 0x0A) : (c - '0'))
 #define HEX_PAIR_TO_BYTE(h, l) ((HEX_CHAR_TO_NIBBLE(h) << 4) + HEX_CHAR_TO_NIBBLE(l))
 
+/* BEGIN TEST */
+
+#ifdef DEBUG
+  #warning Debug mode
+  #define debugPrintLn(...) { if (debugStream) debugStream->println(__VA_ARGS__); }
+  #define debugPrint(...) { if (debugStream) debugStream->print(__VA_ARGS__); }
+#else
+  #define debugPrintLn(...)
+  #define debugPrint(...)
+#endif
+
+/* END TEST */
+
 class TheThingsNetwork
 {
   private:


src/TheThingsNetwork.h
```diff
@@ -23,6 +23,19 @@
 #define HEX_CHAR_TO_NIBBLE(c) ((c >= 'A') ? (c - 'A' + 0x0A) : (c - '0'))
 #define HEX_PAIR_TO_BYTE(h, l) ((HEX_CHAR_TO_NIBBLE(h) << 4) + HEX_CHAR_TO_NIBBLE(l))
 
+/* BEGIN TEST */
```
johanstokking: Remove comments like these

src/TheThingsNetwork.h
```diff
@@ -23,6 +23,19 @@
...
+
+#ifdef DEBUG
```
FokkeZB: So in order to use these in examples I need to do:

```c
#define DEBUG 1
```

*before* I include TheThingsNetwork?

johanstokking: Yes!

FokkeZB: OK, that works and improves the examples a lot not having to repeat this all the time. Two Q's

- Should we expose this via the library (`ttn.debugPrint`) instead of global scope to prevent conflicts?
- Should we use `debugPrintln` (no upper-case `L` since it's `Serial.println` as well)?

Nicolasdejean: I'm going to see if I can put debugPrintLn and debugPrint in the public library.



On Fri, Sep 2, 2016 at 2:32 PM +0200, "Johan Stokking" <notifications@github.com<mailto:notifications@github.com>> wrote:


In src/TheThingsNetwork.h<https://github.com/TheThingsNetwork/arduino-library/pull/12#discussion_r77338168>:

> @@ -23,6 +23,19 @@
>  #define HEX_CHAR_TO_NIBBLE(c) ((c >= 'A') ? (c - 'A' + 0x0A) : (c - '0'))
>  #define HEX_PAIR_TO_BYTE(h, l) ((HEX_CHAR_TO_NIBBLE(h) << 4) + HEX_CHAR_TO_NIBBLE(l))
>
> +/* BEGIN TEST */
> +
> +#ifdef DEBUG


Yes!

-

Nicolasdejean:
It's much more easier to define debugPrintLn and debugPrint. If we put it in the library, we would have to create a function that takes in argument every types of arguments, so void and void *. And in C, according to me, there is no generic variable that combines every type possible.

johanstokking:
Yes true. Also, if they are defined in the C-files, they should not generate compiler errors.

Issue #5, opened 2016-08-19
API: Default to Serial and Serial1


FokkeZB:
The library obviously has to work with any Arduino device, but we can simply sketches for The Things Uno and Node by defaulting to `Serial` for debug and `Serial1` for the modem.

Current:

```c
#define debugSerial Serial
#define loraSerial Serial1

void setup() {
  debugSerial.begin(115200);
  loraSerial.begin(57600);

  ttn.init(loraSerial, debugSerial);

  ttn.showStatus();
}
```

For TT Uno/Node this could be

```c
void setup() {
  ttn.showStatus();
}
```

If you'd need to override for an other device you'd still call `ttn.init()`:

```c
ttn.init(loraSerial, debugSerial);
```

Since it would feel to not use `init()` if you'd use defaults, or if you only need to override one of the two it might be better to dump `init()` all together and have explicit setters:

```c
ttn.setDebugStream(debugSerial);
ttn.setModemStream(debugSerial);
```

FokkeZB:
Maybe `initFor` would be a better name.

@johanstokking am I right there are three "variables" the library needs to know to work on any Arduino:

* The type of LoRaWAN module used (which we internally [map](https://github.com/TheThingsNetwork/arduino-library/blob/master/src/TheThingsNetwork.cpp#L120) to settings)
* The Serial Port the module is on
* The Serial Port used for debug

Since we can [detect the module](https://github.com/TheThingsNetwork/arduino-library/blob/master/src/TheThingsNetwork.cpp#L97-L104) and `Serial` is [always for USB](https://www.arduino.cc/en/Reference/Serial) we only really need to know the Serial Port the LoRaWAN module is on right?

If I'm right then I think we could just remove the `init()` function and add methods to override the Serial Port for the LoRaWAN module in case you don't have it on `Serial1` (or would like to use a different baud rate)

```c
ttn.setModemSerial(modemSerial);
```

This and removing `reset()` will simplify a lot:

```c
#include <TheThingsNetwork.h>

const byte appEui[8] = {0x70, 0xB3, 0xD5, 0x7E, 0xE0, 0xE0, 0x01, 0x4A1};
const byte appKey[16] = {0x73, 0x6D, 0x24, 0xD2, 0x69, 0xBE, 0xE3, 0xAE, 0x0E, 0xCE, 0xF0, 0xBB, 0x6C, 0xA4, 0xBA, 0xFE};

TheThingsNetwork ttn;

[...]

PR #15, opened 2016-09-05
simplify: creating setModemSerial() to define Serials and their Baud rate

Nicolasdejean:
Created initFor(), to simplify examples. Still working on setModemSerial : I need to make sure loraSerial(Serial1) has the right port. #5

### DIFF
--- a/src/TheThingsNetwork.cpp
+++ b/src/TheThingsNetwork.cpp
@@ -4,6 +4,9 @@
 #include <Arduino.h>
 #include <TheThingsNetwork.h>
 
+#define debugSerial Serial
+#define loraSerial Serial
+
 #define debugPrintLn(...) { if (debugStream) debugStream->println(__VA_ARGS__); }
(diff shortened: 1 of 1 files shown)


src/TheThingsNetwork.cpp
```diff
@@ -312,3 +315,19 @@ void TheThingsNetwork::showStatus() {
...
   debugPrintLn(readValue(F("mac get rxdelay2")));
 }
+
+void TheThingsNetwork::setModemSerial(Stream& modemSerial)
+{
+  if (modemSerial.available() == 0)
+    return ;
+}
+
+void TheThingsNetwork::initFor()
```
FokkeZB: I don't think we really need this or `init()`. The constructor of the class could just use `Serial` and `Serial1`(and call `begin()` on them) and then if the user wants to use another Serial Port or Baud Rate he'd call `setModemSerial()`.

FokkeZB:
Closing since #19 does this better

Issue #6, opened 2016-08-19
API: Make reset() private and call it from join() and personalize()


FokkeZB:
All examples call `ttn.reset()` before they call `ttn.join()` or `ttn.personalize()`. If this is a required step, why not make it private and call it from these two methods.

Current:

```c
ttn.reset();
ttn.personalize(devAddr, nwkSKey, appSKey);
```

Result:

```c
ttn.personalize(devAddr, nwkSKey, appSKey);
```

To @johanstokking's note, if the device comes pre-flashed with keys you'd call the methods with no arguments:

```c
while (ttn.join()) {
}
```

The methods should then not call reset() and skip the `mac set` commands.

johanstokking:
Yes, I think we can do the `reset()` in `personalize()` and `join()`. The reason why it's separate, was to be able not to reset and join with the values that are stored in the RN module, instead of providing the values. But we can account for that in `join()` and/or with an overload that doesn't take keys and therefore doesn't reset.

FokkeZB:
Yes, that seems like a good simplification since doing a `reset()` or `personalize()` without `reset()` will be the common scenario. I'll leave this assigned to you as I'm finding my way around in C/Arduino still ;)

FokkeZB:
Of course if we call `reset()` from `join()` this will mean that if it takes a while to do a successful join, every new attempt will first `reset()` the device. This is a bit of an overhead.

johanstokking:
Also, I think we should make it optional not to reset the device at all. I can imagine the AppEUI and AppKey being pre-flashed, especially in The Things Node, that we don't want to reset.

FokkeZB:
Closed by #58

PR #17, opened 2016-09-06
Calling reset in personalize() and join()

Nicolasdejean:
Added a bool as argument to personnalize() and join(), it's true by default.
If you want to do a while with a join(), join() while call reset() the first time than you can put reset = false in the while.
#6


FokkeZB:
I don't know if it is desirable to have another argument?

johanstokking:
I think it's clean, but make it `bool reset = true` in the header file.

PR #19, opened 2016-09-06
Add setters for modem and debug serial and default to Serial1 and Serial0

Nicolasdejean:
created setModemSerial() and setDebugSerial(), containing the serial port by default as first argument and the new serial port chosen by the user (no default new serial port in the prototype).
those functions overrides the last serial port with the new one. #5


src/TheThingsNetwork.cpp
```diff
@@ -312,3 +312,13 @@ void TheThingsNetwork::showStatus() {
   debugPrint(F("RX Delay 2: "));
   debugPrintLn(readValue(F("mac get rxdelay2")));
 }
+
+void TheThingsNetwork::setModemSerial(Stream& modemStream, Stream& replaceSerial)
```
FokkeZB: I'd think you can drop the first argument.

src/TheThingsNetwork.cpp
```diff
@@ -312,3 +312,13 @@ void TheThingsNetwork::showStatus() {
...
+{
+  this->modemStream = &replaceSerial;
+}
+
+void TheThingsNetwork::setDebugSerial(Stream& debugStream, Stream& replaceSerial)
```
FokkeZB: I'd think you can drop the first argument.

src/TheThingsNetwork.h
```diff
@@ -47,6 +47,8 @@ class TheThingsNetwork
     bool join(const byte appEui[8], const byte appKey[16]);
     int sendBytes(const byte* buffer, int length, int port = 1, bool confirm = false);
     int sendString(String message, int port = 1, bool confirm = false);
+    void setModemSerial(Stream& modemStream, Stream& replaceSerial);
```
FokkeZB: I'd think you can drop the first argument.

src/TheThingsNetwork.h
```diff
@@ -47,6 +47,8 @@ class TheThingsNetwork
...
+    void setDebugSerial(Stream& debugStream, Stream& replaceSerial);
```
FokkeZB: I'd think you can drop the first argument.

FokkeZB:
What remains to be added to this PR is defaulting to `Serial` and `Serial1`. I guess this should happen in `init()` where we should:

* Remove the arguments now that we have the setters
* If `this->modemSerial` was not set, default to `Serial1` and call `begin(57600)`
* If `this->debugSerial` was not set, default to `Serial` and call `begin(9600)`

johanstokking:
In the library, we can't reference `Serial1`, because it's only available on some boards. This has to come from the user sketch, as the library is not aware of the selected board in the Arduino IDE.

FokkeZB:
@johanstokking but surely we can default?

johanstokking:
`Serial1` is simply not defined if you have a different board type, e.g. Arduino Uno. So you'll get compilation errors. You can't default to something that's not defined.

FokkeZB:
Still have to get used to compiled languages :S

In that case, with these two setters we might as well remove init() since all it does is set these 2

PS: is the same true for Serial? Or can we default to that for debug still?

src/TheThingsNetwork.cpp
```diff
@@ -7,9 +7,12 @@
 #define debugPrintLn(...) { if (debugStream) debugStream->println(__VA_ARGS__); }
 #define debugPrint(...) { if (debugStream) debugStream->print(__VA_ARGS__); }
 
-void TheThingsNetwork::init(Stream& modemStream, Stream& debugStream) {
-  this->modemStream = &modemStream;
-  this->debugStream = &debugStream;
+void TheThingsNetwork::init(Stream& modemStream) {
+  this->modemStream = &modemStream;
```
FokkeZB: Since we now have setModemStream as well we can remove this argument as well and in init error if modemstream wasn't set already

src/TheThingsNetwork.cpp
```diff
@@ -7,9 +7,12 @@
...
 
-void TheThingsNetwork::init(Stream& modemStream, Stream& debugStream) {
-  this->modemStream = &modemStream;
-  this->debugStream = &debugStream;
+void TheThingsNetwork::init(Stream& modemStream) {
+  this->modemStream = &modemStream; 
+  if (!this->debugStream) {
+    Serial.begin(9600);
+    TheThingsNetwork::setDebugSerial(Serial);
+  }
```
johanstokking: This is assuming that `Serial` is set on all platforms. I can imagine that there are Arduino boards that don't have a `Serial` defined.

Note that if the `debugStream` is not opened yet, this check also returns true. So without initializing the debug serial in the user sketch, it will always be reset to `Serial`, and `begin()` may be called twice.

Also, why calling functions as static functions all the time?

FokkeZB: The [Arduino Reference](https://www.arduino.cc/en/Reference/Serial) does seem to suggest we can assume `Serial` is there:

> All Arduino boards have at least one serial port (also known as a UART or USART): `Serial`. It communicates on digital pins 0 (RX) and 1 (TX) as well as with the computer via USB.

According to the [IfSerial Reference](https://www.arduino.cc/en/Serial/IfSerial) on Leonardo Boards it `Serial` will only be true if it's open, but indeed on others it will always be `true`.

All in all... it looks like there's too many uncertainties and we might be better off just keeping init as it is and have no `set*Serial()`.

@johanstokking let's close then?

johanstokking: Agreed

PR #24, opened 2016-09-09
Calculate size of buffer

### LINKED ISSUE #14: API: Remove length as argument of sendBytes
Current:

```c
```

Simplified:

```c
int sendBytes(const byte* buffer, int port = 1, bool confirm = false);
int sendBytes(const byte* buffer, bool confirm = false);
```

Reasoning:

* Examples have `ttn.sendBytes(payload, sizeof(payload))` so why not do `sizeof(payload)` in the lib and go from 4 to 3 args.
* With an additional version without port you can ask for confirm without having to set port.

FokkeZB:
No longer requires you to pass length, closes #14

Tested by calling:

```c
  byte payload[] = { 0x48, 0x65, 0x6C, 0x6C, 0x6F };
  ttn.sendBytes(payload);
```

Length was correctly calculated as `5`.

### DIFF
--- a/examples/Downlink/Downlink.ino
+++ b/examples/Downlink/Downlink.ino
@@ -33,7 +33,7 @@ void loop() {
   // Send a byte
   byte buf[1];
   buf[0] = 20;
-  int downlinkBytes = ttn.sendBytes(buf, 1);
+  int downlinkBytes = ttn.sendBytes(buf);
 
   if (downlinkBytes > 0) {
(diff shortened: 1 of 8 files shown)


FokkeZB:
Forgive my C ignorance ;)

```
byte payload[] = { 0x00, 0x65, 0x6C, 0x6C, 0x6F };
```

length = 0 because first byte is falsy, as @johanstokking already pointed out:
https://github.com/TheThingsNetwork/arduino-library/issues/14#issuecomment-245288378

Issue #23, opened 2016-09-09
API: Allow user to register function for downlink


FokkeZB:
Receiving a response is not straight forward. You have to send to receive and `sendBytes()` only returns the length. Then you'll have to use `ttn.downlinkPort` and `ttn.downlink` to get the rest of the info.

The Wire library lets you [register a function](https://www.arduino.cc/en/Reference/WireOnReceive) to be called when a message is received. This is a much more friendly API already, although the function still [doesn't receive all info](http://wiring.org.co/reference/libraries/Wire/Wire_onReceive_.html) as arguments, which I think it should.

Current:

```c
void loop() {
  // Send a byte
  byte buf[1];
  buf[0] = 20;
  int downlinkBytes = ttn.sendBytes(buf, 1);

  if (downlinkBytes > 0) {
    debugPrintLn("Received " + String(downlinkBytes) + " bytes on port " + String(ttn.downlinkPort) + ":")

    for (int i = 0; i < downlinkBytes; i++) {
      debugPrint(String(ttn.downlink[i]) + " ");
    }

    debugPrintLn();
  }

  delay(20000);
}
```

Proposed:

```c
void setup() {
  // ..

  ttn.onMessage(receiveMessage);
}

void loop() {
  // send a byte to poll
  var poll = { x00 };
  ttn.sendByes(poll, 1);
}

void receiveEvent(const byte* buffer, int length, int port) {
    debugPrintLn("Received " + String(length) + " bytes on port " + String(port) + ":")

    for (int i = 0; i < downlinkBytes; i++) {
      debugPrint(String(buffer[i]) + " ");
    }

    debugPrintLn();
}
```

Issue #11, opened 2016-08-23
API: sendBytes should return confirmation or error codes


waltertakens:
now sendBytes returns an integer containing the number of bytes received downlinkLength,
 or in case of unconfirmed 0.

Is it easier for the enduser to still return an integer but composed of 2 parts :
msb signed byte  containing the result of the send , eg 0 for ok, -1 for timeout, -2 for busy etc
lsb unsigned byte containing the received downlinkLength

FokkeZB:
@johanstokking can we maybe simplify downlink some while we're at it? I find it weird that you sendBytes in order to receive. I know this is how it technically works, but it's not very intuitive.

**current**

```c
void loop() {
  // Send a byte
  byte buf[1];
  buf[0] = 20;
  int downlinkBytes = ttn.sendBytes(buf, 1);

  if (downlinkBytes > 0) {
    debugPrintLn("Received " + String(downlinkBytes) + " bytes")
    // Print the received bytes
    for (int i = 0; i < downlinkBytes; i++) {
      debugPrint(String(ttn.downlink[i]) + " ");
    }
    debugPrintLn();
  }

  delay(20000);
}
```

**possible simplification**

```c

// let users define a function to pass as callback
void receive(message) {
  // .. do something
}

void loop() {

  // new function handling sendBytes and then looping downlink
  ttn.pingDownlink(receive);

  delay(20000);
}
```

I'm no C/Arduino expert so I don't know if there are limitations that make this impossible or maybe we can even register the callback and an interval in `setup()`? Or is it a single thread?

cambierr:
@FokkeZB It's how it often works in public libs (see Wire lib for instance). User registers a callback function which is then called by the lib when necessary.

FokkeZB:
I've deleted the discussion on an alternative API for receiving messages. Let's do that in #23.

For `sendBytes`, we still need a better response. Currently it returns the number of bytes received in response, but we won't need that anymore with #23.

Instead, it should return a success or error code, e.g.:

- `1` if the [4th `confirm` argument](https://github.com/TheThingsNetwork/arduino-library/blob/master/src/TheThingsNetwork.cpp#L250) was set to to `true` and the message was confirmed indeed.
- `0` if the [4th `confirm` argument](https://github.com/TheThingsNetwork/arduino-library/blob/master/src/TheThingsNetwork.cpp#L250) was not set to `true` and there were no errors.
[...]

PR #34, opened 2016-09-14
Deliver downlink messages to (callback) function

Nicolasdejean:
created onMessage() to clarify sendBytes.
onMessage() contains a buffer downlink and an int downlinkPort to get info on the received message and then send the info to receiveEvent().#23
I couldn't register onMessage() like an event: it needs libraries like <event.h> and <wire.h>, so the user can't choose what to put in onMessage().
I added some (not the right ones) return values for sendBytes.#11

### DIFF
--- a/src/TheThingsNetwork.cpp
+++ b/src/TheThingsNetwork.cpp
@@ -292,14 +292,6 @@ int TheThingsNetwork::onMessage(String response)
     return (1);
 }
 
-int TheThingsNetwork::sendString(String message, int port, bool confirm) {
-  int l = message.length();
-  byte buf[l + 1];
-  message.getBytes(buf, l + 1);
-
-  return sendBytes(buf, l, port, confirm);
-}
-
 void TheThingsNetwork::showStatus() {
   debugPrint(F("EUI: "));
   debugPrintLn(readValue(F("sys get hweui")));
--- a/src/TheThingsNetwork.cpp
+++ b/src/TheThingsNetwork.cpp
@@ -251,33 +251,45 @@ int TheThingsNetwork::sendBytes(const byte* buffer, int length, int port, bool c
   str.concat(port);
   if (!sendCommand(str, buffer, length)) {
     debugPrintLn(F("Send command failed"));
-    return 0;
+    return (-1);
   }
 
   String response = readLine(10000);
   if (response == "") {
     debugPrintLn(F("Time-out"));
-    return 0;
+    return (-2);
   }
   if (response == F("mac_tx_ok")) {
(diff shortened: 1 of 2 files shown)


FokkeZB:
@Nicolasdejean could you give an example of a downlink Sketch? Don't quite get how this would look to the user now, confused by the header `int onMessage(string response)`.

Nicolasdejean:
@FokkeZB The user won't see ttn.onMessage I know it's not user friendly, so I'm trying to see if I can do something so that the user can understand the code that we are giving him.
```
#define DEBUG 1
#include <TheThingsNetwork.h>

// Set your AppEUI and AppKey
const byte appEui[8] = { 0x70, 0xB3, 0xD5, 0x7E, 0xD0, 0x00, 0x0D, 0x35 };
const byte appKey[16] = { 0x34, 0xB8, 0x07, 0x01, 0xAB, 0x87, 0x05, 0x38, 0x34, 0xFB, 0x5C, 0x68, 0x6D, 0x6B, 0xED, 0xF6 };

#define debugSerial Serial
#define loraSerial Serial1

fp_ttn_t fp = 0;
TheThingsNetwork ttn(fp);

void setup()
{
  debugSerial.begin(115200);
  loraSerial.begin(57600);
  ttn.init(loraSerial, debugSerial);
  
  // the device will attempt a join every 10 seconds till the join is successfull
  ttn.join(appEui, appKey, true, 10000, 1000);
  ttn.showStatus();
  debugSerialPrintLn("Setup for The Things Network complete");
}

void loop() {
 ttn.poll(); /* contains sendbytes() that calls onMessage() if it receives a message */
 delay(20000);
}

void receiveEvent(const byte* buffer, int length, int port) /* onMessage() calls receiveEvent() */
{
  debugSerialPrint("Receive" + String(length) + " bytes on port " + String(port) + ":");
  for (int i = 0; i < length; i++)
    debugSerialPrint(String(buffer[i]) + " ");
  debugSerialPrintLn();
}
```

FokkeZB:
I think this is an improvement already since sending and retrieving is now separated and you get all information (buffer, length and port) in one go.

Just to clarify, it's not possible that the user defines what function should be called? For example in `setup()` he would say:

```c
ttn.onMessage(receiveEvent);
```

If that is not possible, then we just have to document `receiveEvent` as convention. I'd suggest to rename it to `message`or `onMessage` to be more aligned to the application libraries.

FokkeZB:
I found a way to let the user define the callback: #43

FokkeZB:
Closing for #43

PR #37, opened 2016-09-14
sendBytes: Return success and error codes

Nicolasdejean:
added return values to sendBytes#11

### DIFF
--- a/src/TheThingsNetwork.cpp
+++ b/src/TheThingsNetwork.cpp
@@ -257,9 +257,11 @@ int TheThingsNetwork::sendBytes(const byte* buffer, int length, int port, bool c
   String response = readLine(10000);
   if (response == "") {
     debugPrintLn(F("Time-out"));
-    if (confirm == true)
+    if (confirm == true) {
+      debugPrintLn(F("Time-out"));
       return 0;
-    return -2;
+    }
+    return 1;
   }
   if (response == F("mac_tx_ok")) {
     debugPrintLn(F("Successful transmission"));
@@ -275,14 +277,12 @@ int TheThingsNetwork::sendBytes(const byte* buffer, int length, int port, bool c
     debugPrint(F("Successful transmission. Received "));
     debugPrint(downlinkLength);
     debugPrintLn(F(" bytes"));
-    if (confirm == true)
-      return 2;
     return 1;
   }
 
   debugPrint(F("Unexpected response: "));
   debugPrintLn(response);
-  return -3;
+  return -10;
 }
 
 int TheThingsNetwork::sendString(String message, int port, bool confirm) {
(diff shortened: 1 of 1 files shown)


FokkeZB:
**NOTE:** This can only be merged after #34 since `sendBytes()` would no longer return the downlink size.

src/TheThingsNetwork.cpp
```diff
@@ -273,11 +275,14 @@ int TheThingsNetwork::sendBytes(const byte* buffer, int length, int port, bool c
     debugPrint(F("Successful transmission. Received "));
     debugPrint(downlinkLength);
     debugPrintLn(F(" bytes"));
-    return downlinkLength;
+    if (confirm == true)
```
FokkeZB: We should simply return 1 if we have a response regardless of `confirm`.

src/TheThingsNetwork.cpp
```diff
@@ -251,17 +251,19 @@ int TheThingsNetwork::sendBytes(const byte* buffer, int length, int port, bool c
...
+    return -1;
   }
 
   String response = readLine(10000);
   if (response == "") {
     debugPrintLn(F("Time-out"));
-    return 0;
+    if (confirm == true)
+      return 0;
+    return -2;
```
FokkeZB: If we didn't ask for a confirm we should return `1` here since there's no error.

src/TheThingsNetwork.cpp
```diff
@@ -251,17 +251,19 @@ int TheThingsNetwork::sendBytes(const byte* buffer, int length, int port, bool c
   str.concat(port);
   if (!sendCommand(str, buffer, length)) {
     debugPrintLn(F("Send command failed"));
-    return 0;
+    return -1;
   }
 
   String response = readLine(10000);
   if (response == "") {
     debugPrintLn(F("Time-out"));
```
FokkeZB: Not sure if we should always print this debug line or only when we asked for a confirm

src/TheThingsNetwork.cpp
```diff
@@ -273,11 +275,14 @@ int TheThingsNetwork::sendBytes(const byte* buffer, int length, int port, bool c
...
     debugPrintLn(F(" bytes"));
-    return downlinkLength;
+    if (confirm == true)
+      return 2;
+    return 1;
   }
 
   debugPrint(F("Unexpected response: "));
   debugPrintLn(response);
+  return -3;
```
FokkeZB: Let's make this `-10` so that we have space to introduce more (known) error codes.

src/TheThingsNetwork.cpp
Nicolasdejean: It's a time-out, wouldn't we want a different return for that ?

FokkeZB: That depends. I'm not sure if without confirm you'd also get this empty response or if it's indeed a separate error. @johanstokking?

src/TheThingsNetwork.cpp
johanstokking: Empty response is always error `-2`

Issue #32, opened 2016-09-09
Initialize library for specific frequency plan


johanstokking:
Issues #21 and #22 are frequency plan specific, but the library does not accommodate for frequency plans yet. Instead, it only checks the RN module. We cannot assume that RN2483 is used in EU868.

We should define frequency plans, e.g. `#define TTN_EU868 1`, and initialize the library with this frequency plan. Instead of checking the RN module, it should assert that the RN module matches the frequency plan (e.g. `TTN_EU868` needs `RN2483`), and based on the frequency plan, configure the channels and duty cycle.

FokkeZB:
Closed by #63

PR #38, opened 2016-09-14
initialize the library with a frequency plan

Nicolasdejean:
Initialize the library to TTN_EU868. #32


src/TheThingsNetwork.h
```diff
@@ -41,6 +46,7 @@ class TheThingsNetwork
   public:
     int downlinkPort;
     byte downlink[64];
+    TheThingsNetwork(fp_ttn_t fp);
```
FokkeZB: Isn't it confusing that we'd now have both a constructor argument and a separate `init()`? Can't we add it to `init()` or move all of them to individual setters so that `init()` doesn't grow into a big monster?

src/TheThingsNetwork.h
```diff
@@ -44,6 +44,13 @@ class TheThingsNetwork
     bool enableFsbChannels(int fsb);
 
   public:
+    TheThingsNetwork(fp_ttn_t *fp)
+    {
+      #ifdef TTN_FP_EU868
+        *fp = 1;
+      #endif
+      this->fp = *fp;
+    }
```
johanstokking: The user provides the `fp`, so just get `fp_ttn_t fp` from the user and set it as `this->fp`. No pointers, no define checks. We can't define stuff that the library uses, unfortunately, that's not supported in Arduino.

src/TheThingsNetwork.h
```diff
@@ -39,6 +44,10 @@ class TheThingsNetwork
     bool enableFsbChannels(int fsb);
 
   public:
+    TheThingsNetwork(fp_ttn_t fp)
+    {
+      this->fp = fp;
+    }
```
johanstokking: Initialize in cpp file

src/TheThingsNetwork.cpp
```diff
@@ -312,3 +312,55 @@ void TheThingsNetwork::showStatus() {
...
+       str.concat(F("mac set ch status "));
+       str.concat(ch);
+       str.concat(F(" on"));
+       sendCommand(str);
+       str = "";
+       freq = freq + 200000;
+   }
+}
+
+void TheThingsNetwork::set_dcycle() {
```
johanstokking: Switch/case for `fp` and ask @tftelkamp for settings per frequency plan. Start with EU868 and US915

src/TheThingsNetwork.cpp
```diff
@@ -313,54 +313,55 @@ void TheThingsNetwork::showStatus() {
...
     str = "";
   }
+  str.concat(F("mac set ch drrange 1 0 6"));
+  sendCommand(str);
+  str = "";
 }
 
+void TheThingsNetwork::configure_channels() {
+  switch (this->fp) {
+      case 1:
```
johanstokking: - Check spacing (2 vs 4)
- `case TTN_FP_EU868`

src/TheThingsNetwork.h
```diff
@@ -44,21 +45,18 @@ class TheThingsNetwork
...
     void reset(bool adr = true, int sf = DEFAULT_SF, int fsb = DEFAULT_FSB);
     bool personalize(const byte devAddr[4], const byte nwkSKey[16], const byte appSKey[16]);
     bool join(const byte appEui[8], const byte appKey[16]);
     int sendBytes(const byte* buffer, int length, int port = 1, bool confirm = false);
     int sendString(String message, int port = 1, bool confirm = false);
     void showStatus();
-    void configure_channels(int total_ch);
-    void set_dcycle();
+    void configure_channels();
+    void configure_EU868();
```
johanstokking: This is not public, that's the point of passing a frequency plan, we hide it from the user

src/TheThingsNetwork.h
```diff
@@ -56,7 +57,6 @@ class TheThingsNetwork
     int sendString(String message, int port = 1, bool confirm = false);
     void showStatus();
     void configure_channels();
```
johanstokking: Same for channels; this is FP specific

johanstokking:
- [ ] Include US915
- [ ] One method for frequency plan that sets power index, data rate (move from `reset()`), frequencies, channels etc

src/TheThingsNetwork.cpp
```diff
@@ -312,13 +313,19 @@ void TheThingsNetwork::showStatus() {
...
+  debugPrintLn(readValue(F("mac get ch freq 3")));
+
 }
 
 void TheThingsNetwork::configure_EU868() {
   int ch;
   long int freq = 867100000;
   String str = "";
 
+  debugPrintLn(F("wrong conf"));
```
johanstokking: Remove

src/TheThingsNetwork.cpp
```diff
@@ -312,13 +313,19 @@ void TheThingsNetwork::showStatus() {
   debugPrintLn(readValue(F("mac get rxdelay1")));
   debugPrint(F("RX Delay 2: "));
   debugPrintLn(readValue(F("mac get rxdelay2")));
+  debugPrint(F("ch 2 US : "));
+  debugPrintLn(readValue(F("mac get ch freq 2")));
+  debugPrint(F("ch 3 US : "));
+  debugPrintLn(readValue(F("mac get ch freq 3")));
```
johanstokking: This is not applicable on all frequency plans

src/TheThingsNetwork.cpp
```diff
@@ -360,14 +354,13 @@ void TheThingsNetwork::configure_EU868() {
   str = "";
 }
 
-void TheThingsNetwork::configure_US915()
-{
+void TheThingsNetwork::configure_US915() {
   int ch;
   long int freq = 903900000;
   String str = "";
```
johanstokking: Remove second empty line

src/TheThingsNetwork.cpp
```diff
@@ -356,32 +356,22 @@ void TheThingsNetwork::configure_EU868() {
...
     str.concat(ch);
-    str.concat(F(" 0 5"));
+    str.concat(F(" 0 3"));
     sendCommand(str);
     str = "";
     str.concat(F("mac set ch status "));
     str.concat(ch);
     str.concat(F(" on"));
     sendCommand(str);
     str = "";
```
johanstokking: Look at the original code. There are 64 + 8 channels for US915.

johanstokking: Also, whether of not to on/off a channel depends on the FSB. See original code

Nicolasdejean:
@johanstokking ok I didn't see the enableFsbChannels()... Do I also move the dr switch/case from reset() to configure_channels() ?

API.md
```diff
@@ -0,0 +1,121 @@
+# API Reference
+Include and instantiate the TheThingsNetwork class.
+
+```c
+#include <TheThingsNetwork.h>
+TheThingsNetwork ttn
+```
+
+## Method: init
+Initialise the library with the streams it should communicate with.
```
johanstokking: Initialize; en-US everywhere

johanstokking:
@Nicolasdejean let's keep `reset()` for resetting ABP and OTAA settings, and configure everything related to the frequency plans in `init()`: data rates, channels, duty cycles, etc.

src/TheThingsNetwork.cpp
```diff
@@ -411,6 +340,78 @@ void TheThingsNetwork::configure_channels() {
...
+    return;
+  }
+  model = version.substring(0, version.indexOf(' '));
+  debugPrint(F("Version is "));
+  debugPrint(version);
+  debugPrint(F(", model is "));
+  debugPrintLn(model);
+  int dr = -1;
+  String str = "";
+  if (model == F("RN2483")) {
```
johanstokking: Don't check on the `model`. We know what to configure based on the frequency plan

src/TheThingsNetwork.cpp
```diff
@@ -411,6 +340,78 @@ void TheThingsNetwork::configure_channels() {
...
+   debugPrintLn("PassingThrought");
```
johanstokking: Remove

Nicolasdejean:
@johanstokking I enable the channels in a loop for EU in configure_EU868 or by calling enableFsbChannels() in configure() for US in configure. Is there an enable loop I forgot ?

src/TheThingsNetwork.cpp
```diff
@@ -312,107 +313,97 @@ void TheThingsNetwork::configure_EU868() {
...
   int ch;
+  int dr = -1;
   String str = "";
  
   sendCommand(F("radio set freq 904200000"));
   str = "";
+  str.concat(F("mac set pwridx "));
+  str.concat(TTN_PWRIDX_915);
+  sendCommand(str);
+  enableFsbChannels(fsb);
```
johanstokking: Inline `enableFsbChannels()` for now; it's too US915 specific.

src/TheThingsNetwork.cpp
```diff
@@ -312,107 +313,97 @@ void TheThingsNetwork::configure_EU868() {
...
 
 void TheThingsNetwork::configure_channels(int sf, int fsb) {
   switch (this->fp) {
     case TTN_FP_EU868:
-      configure_EU868();
+      configure_EU868(sf);
       break;
     case TTN_FP_US915:
-      configure_US915();
+      configure_US915(sf, fsb);
```
johanstokking: Yep this is good

johanstokking:
@FokkeZB @Nicolasdejean why is this PR so big? Should it be rebased?

FokkeZB:
I guess, this has a lot of commits already in a long time.. 48 commits?! It should be rebased or maybe just re-done one a new branch of off master.

PR #39, opened 2016-09-14
Handle recommended reset() before while(join()) internally

Nicolasdejean:
with three more parameters: reset (true or false)#6 , number of delay and max attempts. No need to use a while(join()).


src/TheThingsNetwork.cpp
```diff
@@ -219,7 +219,16 @@ bool TheThingsNetwork::personalize(const byte devAddr[4], const byte nwkSKey[16]
...
 
-bool TheThingsNetwork::join(const byte appEui[8], const byte appKey[16]) {
+bool TheThingsNetwork::join(const byte appEui[8], const byte appKey[16], bool reset, long int nbr_delay, int max_attempts) {
+  
+  static int nbr_attempts;
+
+  if (nbr_attempts++ <= max_attempts)
+     if (TheThingsNetwork::join(appEui, appKey, true, nbr_delay, max_attempts) == true)
+        return (true);
+  delay(nbr_delay);
```
johanstokking: Why not loop here?

src/TheThingsNetwork.cpp
```diff
@@ -219,7 +219,16 @@ bool TheThingsNetwork::personalize(const byte devAddr[4], const byte nwkSKey[16]
...
+  if (reset == true)
+    TheThingsNetwork::reset();
```
johanstokking: See other comment

src/TheThingsNetwork.h
```diff
@@ -44,7 +44,7 @@ class TheThingsNetwork
     void init(Stream& modemStream, Stream& debugStream);
     void reset(bool adr = true, int sf = DEFAULT_SF, int fsb = DEFAULT_FSB);
     bool personalize(const byte devAddr[4], const byte nwkSKey[16], const byte appSKey[16]);
-    bool join(const byte appEui[8], const byte appKey[16]);
+    bool join(const byte appEui[8], const byte appKey[16], bool reset = true, long int nbr_delay = 20000, int max_attempts = 1000);
```
FokkeZB: - I would bring the delay down to `10000`.
- I would default `max_attempts` to `0` meaning it will try forever, which is what our examples currently do and the node is useless without join.

Nicolasdejean:
@johanstokking Do you mean loop after delay() ?

johanstokking:
I mean, loop in `join()`, until either a successful join or the number of attempts is reached.

Nicolasdejean:
Well there's already a loop in join(): join() recalls himself recursively until it is a successfull join() or the number of attempts is reached.

johanstokking:
@Nicolasdejean I see that, but this recursive pattern consumes an unnecessary amount of memory on the stack and is hard to read, while the only thing you want to do here is loop.

Also don't call `join()` as static method.

src/TheThingsNetwork.h
Nicolasdejean: I default max_attempts to -1 so it could loop forever until a successful join.

FokkeZB: Defaulting to `0` also works fine since you start at 1 and use `!=` not `<=`

src/TheThingsNetwork.cpp
```diff
@@ -219,13 +219,14 @@ bool TheThingsNetwork::personalize(const byte devAddr[4], const byte nwkSKey[16]
   return true;
 }
 
-bool TheThingsNetwork::join(const byte appEui[8], const byte appKey[16], long int nbr_delay, int max_attempts) {
+bool TheThingsNetwork::join(const byte appEui[8], const byte appKey[16], bool restart, long int nbr_delay, int max_attempts) {
```
FokkeZB: We shouldn't add this and just do the same as in #42. Add a proto without the first two args that does no reset and in this one do reset and set the appeui and appkey

src/TheThingsNetwork.cpp
```diff
@@ -219,14 +219,40 @@ bool TheThingsNetwork::personalize(const byte devAddr[4], const byte nwkSKey[16]
...
+  return false;
+}
+
+
+bool TheThingsNetwork::join(const byte appEui[8], const byte appKey[16], long int nbr_delay, int max_attempts) {
+  int nbr_attempts = 0;
+
+  while (nbr_attempts++ != max_attempts) {
+    delay(nbr_delay);
+    reset();
```
FokkeZB: I think you should only call reset() once

Nicolasdejean: If I call reset() only once, join() fails all the time for me on the second or third loop. It says there are no free channels available... Is that normal ?
```
Sending: mac set appeui with 8 bytes
Sending: mac set deveui 0004A30B001B4C9F
Sending: mac set appkey with 16 bytes
Sending: mac join otaa
Response is not OK: no_free_ch
Send join command failed
```

src/TheThingsNetwork.cpp
```diff
@@ -219,29 +219,61 @@ bool TheThingsNetwork::personalize(const byte devAddr[4], const byte nwkSKey[16]
...
+}
 
-  debugPrint(F("Join accepted. Status: "));
-  debugPrintLn(readValue(F("mac get status")));
-  return true;
+bool TheThingsNetwork::join(const byte appEui[8], const byte appKey[16], long int nbr_delay, int max_attempts) {
+  int nbr_attempts = 0;
+
+  reset();
+  while (nbr_attempts++ != max_attempts) {
```
FokkeZB: I think you could call `join()` (no args) here if you'd move out `mac set appeui` and `mac set appkey` out of the while. I'm not sure (@johanstokking?) but I think you don't need to do this before every attempt to `mac join otaa`?

FokkeZB: Ah, [indeed](https://github.com/TheThingsNetwork/arduino-device-lib/pull/39#discussion_r79361578)

src/TheThingsNetwork.cpp
```diff
@@ -219,29 +219,38 @@ bool TheThingsNetwork::personalize(const byte devAddr[4], const byte nwkSKey[16]
...
-    debugPrintLn(F("Send join command failed"));
-    return false;
-  }
-
-  String response = readLine(10000);
-  if (response != F("accepted")) {
-    debugPrint(F("Join not accepted: "));
-    debugPrintLn(response);
-    return false;
+  while (nbr_attempts++ != retries) {
```
johanstokking:
```
while (retries != 0) {
  if (retries > 0) {
    retries--;
  }
}
```

Comment with support for infinite retries (`-1`)

FokkeZB:
@Nicolasdejean also make `reset()` private now both `personalize()` and `join()` handle this internally, so we can close #6

src/TheThingsNetwork.cpp
```diff
@@ -220,15 +220,16 @@ bool TheThingsNetwork::personalize(const byte devAddr[4], const byte nwkSKey[16]
...
-  int nbr_attempts = 0;
-
   String devEui = readValue(F("sys get hweui"));
   String str = "";
   str.concat(F("mac set deveui "));
   str.concat(devEui);
   sendCommand(str);
-  while (nbr_attempts++ != retries) {
+  while (retries != 0) {
     delay(retry_delay);
```
johanstokking: Do not start with a delay. Delay at the end, when join didn't succeed.

src/TheThingsNetwork.cpp
```diff
@@ -219,29 +219,39 @@ bool TheThingsNetwork::personalize(const byte devAddr[4], const byte nwkSKey[16]
   return true;
 }
 
-bool TheThingsNetwork::join(const byte appEui[8], const byte appKey[16]) {
+bool TheThingsNetwork::join(int retries, long int retry_delay) {
   String devEui = readValue(F("sys get hweui"));
-  sendCommand(F("mac set appeui"), appEui, 8);
   String str = "";
   str.concat(F("mac set deveui "));
   str.concat(devEui);
```
johanstokking: Put L223-226 also in `reset`

src/TheThingsNetwork.h
```diff
@@ -37,14 +37,15 @@ class TheThingsNetwork
...
 
   public:
     int downlinkPort;
     byte downlink[64];
     void init(Stream& modemStream, Stream& debugStream);
-    void reset(bool adr = true, int sf = DEFAULT_SF, int fsb = DEFAULT_FSB);
     bool personalize(const byte devAddr[4], const byte nwkSKey[16], const byte appSKey[16]);
-    bool join(const byte appEui[8], const byte appKey[16]);
+    bool join(int retries = -1, long int retry_delay = 10000);
+    bool join(const byte appEui[8], const byte appKey[16], int retries = -1, long int retry_delay = 10000);
```
johanstokking: We use *lowerCamelCase* for parameters, so make it `retryDelay`, just like `appKey`

FokkeZB:
@johanstokking conflict resolved, ready to merge

PR #41, opened 2016-09-14
defined debugSerialPrint/Ln for Serial and debugPrint/Ln for Stream

Nicolasdejean:
to avoid warnings while compilating so in examples use debugSerialPrint/Ln and debugPrint/Ln in library.
You have to: #define DEBUG 1 in examples #4

### DIFF
--- a/src/TheThingsNetwork.h
+++ b/src/TheThingsNetwork.h
@@ -23,6 +23,17 @@
 #define HEX_CHAR_TO_NIBBLE(c) ((c >= 'A') ? (c - 'A' + 0x0A) : (c - '0'))
 #define HEX_PAIR_TO_BYTE(h, l) ((HEX_CHAR_TO_NIBBLE(h) << 4) + HEX_CHAR_TO_NIBBLE(l))
 
+#ifdef DEBUG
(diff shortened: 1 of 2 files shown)


johanstokking:
This defines it in the header so it increases the likelihood of define collisions no?

src/TheThingsNetwork.h
```diff
@@ -23,6 +23,15 @@
 #define HEX_CHAR_TO_NIBBLE(c) ((c >= 'A') ? (c - 'A' + 0x0A) : (c - '0'))
 #define HEX_PAIR_TO_BYTE(h, l) ((HEX_CHAR_TO_NIBBLE(h) << 4) + HEX_CHAR_TO_NIBBLE(l))
 
+#ifdef DEBUG
+#define debugPrintLn(...) { if (debugSerial) debugSerial.println(__VA_ARGS__); }
+#define debugPrint(...) { if (debugSerial) debugSerial.print(__VA_ARGS__); }
+#else
+#define debugPrintLn(...)
+#define debugPrint(...)
+#endif
```
johanstokking: No need to put it in the header

johanstokking:
This PR is effectively an empty new line now, isn't it?

PR #42, opened 2016-09-14
called reset() in personalize()

Nicolasdejean:
Added a bool in parameter: true does a reset, false doesn't.#6


src/TheThingsNetwork.cpp
```diff
@@ -201,7 +201,10 @@ bool TheThingsNetwork::enableFsbChannels(int fsb) {
   return true;
 }
 
-bool TheThingsNetwork::personalize(const byte devAddr[4], const byte nwkSKey[16], const byte appSKey[16]) {
+bool TheThingsNetwork::personalize(const byte devAddr[4], const byte nwkSKey[16], const byte appSKey[16], bool reset) {
+  
+  if (reset == true)
+    TheThingsNetwork::reset();
```
johanstokking:
```
if reset() {
  reset();
}
```

FokkeZB:
@johanstokking can't we just call `reset()` unless `personalize()` was called with no arguments? Or am I not understanding the reason for `reset()`?

johanstokking:
Yep. I think if you want to personalize with custom keys, that it's good to reset first. Otherwise, when you want to personalize with whatever is in the RN module, you don't reset first.

src/TheThingsNetwork.cpp
```diff
@@ -201,7 +201,11 @@ bool TheThingsNetwork::enableFsbChannels(int fsb) {
   return true;
 }
 
-bool TheThingsNetwork::personalize(const byte devAddr[4], const byte nwkSKey[16], const byte appSKey[16]) {
+bool TheThingsNetwork::personalize(const byte devAddr[4], const byte nwkSKey[16], const byte appSKey[16], bool reset) {
```
FokkeZB: As per @johanstokking's comment https://github.com/TheThingsNetwork/arduino-device-lib/pull/42#issuecomment-247356579 we can drop the `reset` arg and add a `personalize()` proto with no args which does no `reset()` and skips the 3 `mac set` commands. Of course this means the `personalize()` which does take 3 arguments can start with `reset()` and the 3 commands and then call the `personalize()` with no arguments to keep code DRY.

src/TheThingsNetwork.cpp
```diff
@@ -201,16 +201,15 @@ bool TheThingsNetwork::enableFsbChannels(int fsb) {
   return true;
 }
 
-bool TheThingsNetwork::personalize(const byte devAddr[4], const byte nwkSKey[16], const byte appSKey[16], bool reset) {
+bool TheThingsNetwork::personalize(const byte devAddr[4], const byte nwkSKey[16], const byte appSKey[16]) {
   
-  if (reset) {
+  if (devAddr != NULL && nwkSKey != NULL && appSKey != NULL) {
```
FokkeZB: @johanstokking would know better, but I think it's preferable to have a separate proto with no args. That makes it more clear that you either need to pass all 3 or none.

src/TheThingsNetwork.cpp
```diff
@@ -203,12 +203,15 @@ bool TheThingsNetwork::enableFsbChannels(int fsb) {
 
 bool TheThingsNetwork::personalize(const byte devAddr[4], const byte nwkSKey[16], const byte appSKey[16]) {
   
-  if (devAddr != NULL && nwkSKey != NULL && appSKey != NULL) {
-    TheThingsNetwork::reset();
-    sendCommand(F("mac set devaddr"), devAddr, 4);
-    sendCommand(F("mac set nwkskey"), nwkSKey, 16);
-    sendCommand(F("mac set appskey"), appSKey, 16);
-  }
+  TheThingsNetwork::reset();
```
johanstokking: Why static?

src/TheThingsNetwork.cpp
```diff
@@ -203,12 +203,15 @@ bool TheThingsNetwork::enableFsbChannels(int fsb) {
...
+  sendCommand(F("mac set devaddr"), devAddr, 4);
+  sendCommand(F("mac set nwkskey"), nwkSKey, 16);
+  sendCommand(F("mac set appskey"), appSKey, 16);
+  return (personalize());
```
johanstokking: Don't use parentheses when they're not necessary

johanstokking:
@Nicolasdejean no it's OK like this

Code is approved

@FokkeZB can you test?

FokkeZB:
I've successfully tested with this in `setup()`

```c
  ttn.init(loraSerial, debugSerial);

  debugPrintLn("Personalize with keys");
  ttn.personalize(devAddr, nwkSKey, appSKey);

  debugPrintLn("Send 0x11");
  byte payload[] = { 0x11 };
  ttn.sendBytes(payload, sizeof(payload));

  debugPrintLn("Personalize without keys");
  ttn.personalize();

  debugPrintLn("Send 0x22");
  byte payload2[] = { 0x22 };
  ttn.sendBytes(payload2, sizeof(payload2));
```

Both messages came through fine.

PR #45, opened 2016-09-16
wait process in ttn.init()

### LINKED ISSUE #44: Hide wait process in setup
```c
  // Wait a maximum of 10s for Serial Monitor
  while (!debugSerial && millis() < 10000);
```

I think this line can be put in the ttn.init() function.
If it create a problem, we can add a third parameter to init to disable the wait.
We can also create a separate function in ttn to do so.

Nicolasdejean:
Put the line:
``` while (!&debugStream && millis() < time) ```
with "time" an long int in parameter (if time  = 0; it disable the wait proccess).

Closes #44


FokkeZB:
On second thought, I'm not sure if we should do #44 at all. It makes it feel like a feature of the lib, while it's just an Arduino debug best practice.

johanstokking:
Agreed

Issue #53, opened 2016-09-22
After successful join() debugPrint DevAddr


FokkeZB:
Since it will change after join: https://github.com/TheThingsNetwork/arduino-device-lib/pull/46#discussion_r79986695

Nicolasdejean:
closed by #60

PR #60, opened 2016-09-22
show the devAddr after join

Nicolasdejean:
deleted it in showStatus() #53


FokkeZB:
@Nicolasdejean next time please update [API.md](https://github.com/TheThingsNetwork/arduino-device-lib/blob/master/API.md#method-showstatus) when you make API changes.

Issue #30, opened 2016-09-09
Track airtime used per day


tftelkamp:
TTN has a 30 second fair access policy. The nodes should be able to keep track of their daily airtime usage to ensure they don't violate it. 

It would be nice if we implement this in the arduino-library, as an example for people how to track this.

The node can also make decisions (e.g. spreading factor) based on remaining airtime.

cambierr:
Afaik, the uno does not have a clock nor static memory which means the only way to do this would be to track that time since the last boot, and reset in when > 1day.

Since most of the devices that will use this library are used for workshops / home dev / geek stuff, they will probably be rebooted very often... meaning that the time tracking won't work. Am I wrong ?

tftelkamp:
It's about making users aware of the airtime usage, and giving them an example of how to track it. Look at it as education. That you lose track on the Uno after a reboot is not so much a concern.

cambierr:
Ho sorry, I thought you wanted to implement a local airtime controller

PR #65, opened 2016-09-27
Track airTime

Nicolasdejean:
Added a timer in sendbytes that calculates air-time of the uplink and a function showAirTime() to show you your air-time. #30


src/TheThingsNetwork.h
```diff
@@ -43,6 +44,7 @@ class TheThingsNetwork
   public:
     void init(Stream& modemStream, Stream& debugStream);
     void showStatus();
+    void showAirTime();
```
FokkeZB: I think we should only have a property or getter to get it, not print it

src/TheThingsNetwork.cpp
```diff
@@ -277,15 +277,18 @@ bool TheThingsNetwork::join(const byte appEui[8], const byte appKey[16], int ret
...
   str.concat(confirm ? F("cnf ") : F("uncnf "));
   str.concat(port);
   if (!sendCommand(str, payload, length)) {
     debugPrintLn(F("Send command failed"));
+    this->timer = this->timer + millis() - start;
     return -1;
   }
-
+  
+  this->timer = this->timer + millis() - start;
```
FokkeZB: Isn't the airtime something we can calculate based on payload size and channel? Timing it like this will include the delay waiting for response (currently 2, will be 5)

src/TheThingsNetwork.cpp
```diff
@@ -324,9 +324,20 @@ int TheThingsNetwork::poll(int port, bool confirm) {
...
+  int band = 125;
+
+  float Tsym = pow(2, sf) / band;
+  float Tpreamble = (ps + 4.25) * Tsym;
+  unsigned int payLoadSymbNb = 8 + (max(ceil((8 * payloadSize - 4 * sf + 28 + 16 - 20 * (0))/ (4 * (sf - 2 * 1))) * (4), 0));
+  float Tpayload = payLoadSymbNb * Tsym;
+  float Tpacket = Tpreamble + Tpayload;
+  this->AirTimePerPackets = Tpacket;
+  this->messDay = (floor(30000 / Tpacket));
+  this->messHour = (30000 / 24 / Tpacket);
```
johanstokking: This is testing code no?

Nicolasdejean: @johanstokking messDay and messHour are just tests, I forgot to delete them... but the rest is there to calculate the airtime:  ``` this->AirTimePerPackets = Tpacket; ```

src/TheThingsNetwork.cpp
```diff
@@ -321,6 +321,20 @@ int TheThingsNetwork::poll(int port, bool confirm) {
   return sendBytes(payload, 1, port, confirm);
 }
 
+void TheThingsNetwork::airTimeValue(int payloadSize) {
```
johanstokking: Rename to `trackAirtime`

src/TheThingsNetwork.cpp
```diff
@@ -321,6 +321,20 @@ int TheThingsNetwork::poll(int port, bool confirm) {
...
+  payloadSize = 13 + payloadSize;
+  int sf = TTN_DEFAULT_SF;
```
johanstokking: Use actual `sf`

src/TheThingsNetwork.cpp
```diff
@@ -321,6 +321,20 @@ int TheThingsNetwork::poll(int port, bool confirm) {
...
+  int ps = 8;
+  int band = 125;
```
johanstokking: This is frequency plan specific

src/TheThingsNetwork.cpp
```diff
@@ -321,6 +321,20 @@ int TheThingsNetwork::poll(int port, bool confirm) {
...
+
+  float Tsym = pow(2, sf) / band;
+  float Tpreamble = (ps + 4.25) * Tsym;
+  unsigned int payLoadSymbNb = 8 + (max(ceil((8 * payloadSize - 4 * sf + 28 + 16 - 20 * (0))/ (4 * (sf - 2 * 1))) * (4), 0));
+  float Tpayload = payLoadSymbNb * Tsym;
+  float Tpacket = Tpreamble + Tpayload;
+  this->AirTimePerPackets = (Tpacket / 10);
```
johanstokking: Add this to private field `airtime` (note casing)

src/TheThingsNetwork.h
```diff
@@ -43,6 +44,7 @@ class TheThingsNetwork
   public:
     void init(Stream& modemStream, Stream& debugStream);
     void showStatus();
+    void airTimeValue(int payloadSize);
```
johanstokking: This is not public. Call internally when sending data

src/TheThingsNetwork.cpp
```diff
@@ -321,18 +320,42 @@ int TheThingsNetwork::poll(int port, bool confirm) {
...
+void TheThingsNetwork::trackAirtime(int payloadSize) {
   payloadSize = 13 + payloadSize;
-  int sf = TTN_DEFAULT_SF;
-  int ps = 8;
-  int band = 125;
+  int sf = getInfo(readValue(F("radio get sf")));
+  int ps = getInfo(readValue(F("radio get prlen")));
+  int band = getInfo(readValue(F("radio get bw")));
+  int header = getInfo(readValue(F("radio get crc")));
+  int cr = getInfo(readValue(F("radio get cr")));
```
johanstokking: These values do not change all the time right? I think it's more power efficient to read these values once and use the variables instead of reading them out of the RN module

src/TheThingsNetwork.cpp
```diff
@@ -321,18 +320,42 @@ int TheThingsNetwork::poll(int port, bool confirm) {
...
+  sf >= 11 ? de = 1 : de = 0;
 
   float Tsym = pow(2, sf) / band;
   float Tpreamble = (ps + 4.25) * Tsym;
-  unsigned int payLoadSymbNb = 8 + (max(ceil((8 * payloadSize - 4 * sf + 28 + 16 - 20 * (0))/ (4 * (sf - 2 * 1))) * (4), 0));
+  unsigned int payLoadSymbNb = 8 + (max(ceil((8 * payloadSize - 4 * sf + 28 + 16 - 20 * header) / (4 * (sf - 2 * de))) * (cr + 4), 0));
   float Tpayload = payLoadSymbNb * Tsym;
   float Tpacket = Tpreamble + Tpayload;
-  this->AirTimePerPackets = (Tpacket / 10);
+  this->airtime = (Tpacket / 10);
```
johanstokking: Add to the `airtime`

FokkeZB:
@Nicolasdejean could you resolve the conflict and document what the public API/use is of this change and what I should test?

Nicolasdejean:
@FokkeZB In API.md I put the use of trackAirtime in sendBytes since its a private function. Also @johanstokking shouldn't we print the airtime in the examples to let the user be aware of the 30 seconds fair policy.

FokkeZB:
If we don't debug print nor make it public it's of no use.

johanstokking:
Yep, true. Put it in `showStatus()`

Nicolasdejean:
@johanstokking @FokkeZB  It's going to change everytime we send a message so shouldn't we print it  in ``` sendBytes() ``` ? Or in another function in ``` void loop ``` ?

johanstokking:
Yes, we can too

FokkeZB:
Isn't this information users might want to calculate with? What if we make the property public so that you can read it and do whatever you like (reset every day perhaps)? We could still also include it in showStatus (total airtime) log in sendBytes (added and total airtime)

FokkeZB:
Also, don't the other commands like for join also cost air time?

Nicolasdejean:
I might be wrong but doesn't the 30 seconds start when we successfully joined the ttn network ?

Nicolasdejean:
@FokkeZB @johanstokking  Should I put a limitation to sendbytes(), if the airtime reaches 30 s I should stop the uplink and only do downlinks until I receive 10 messages (I read that it was 10 messages downlink per day) ?

Nicolasdejean:
@FokkeZB I printed the airtime in showstatus. The formula I used apply's to all SF, changing it will, I think, be no use to the user, and the airtime is going to reset everytime you close the workshop so it think it's better to let ``` trackAirtime() ``` private. What I could do is put ``` fillAirtimeInfo() ``` public so that the user may change the values and may fool around with the airtime system.

johanstokking:
@Nicolasdejean no blocking, also no fooling around by the user. It's just an indication on the node-side, as @FokkeZB said to change behavior. The tracking itself will be done in the Network Server, further upstream after the Router and Broker.

johanstokking:
@FokkeZB can you test and merge?

FokkeZB:
I will, but the branch has conflicts and is all dirty. It should be rebased and cherry picked.

johanstokking:
@Nicolasdejean please resolve conflicts

johanstokking:
Replaced by #84

PR #82, opened 2016-10-07
Added SoftwareSerial listen function

bengchet:
I tweaked the library a bit to enable listen function if users use this library for SoftwareSerial case. In original library, SoftwareSerial can be used but if the user use another SoftwareSerial class in the same sketch, the library might not work. In this tweaked library, I added some lines to check if user uses modemStream as HardwareSerial or SoftwareSerial, if SoftwareSerial, modemStream will intend to listen every time before sendCommand function is executed.

### DIFF
--- a/src/TheThingsNetwork.cpp
+++ b/src/TheThingsNetwork.cpp
@@ -67,6 +67,10 @@ char btohexa_low(unsigned char b) {
 }
 
 bool TheThingsNetwork::sendCommand(String cmd, const byte *buf, int length, int waitTime) {
+
+  // listen if softwareserial is used
+  if(softser){softser->listen();}
+
   debugPrint(F("Sending: "));
   debugPrint(cmd);
   debugPrint(F(" with "));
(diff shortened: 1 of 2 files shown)


johanstokking:
The reason why we're using `Stream` is because you can use `SoftwareSerial`, `HardwareSerial`, `AltSoftSerial`, etc. What functionality do you specifically miss in the current implementation?

bengchet:
In SoftwareSerial, the most important point is listen() function, which will make specific port to listen to incoming data at RX. This case is important when there are more than 1 SoftwareSerial is implemented.

For example, a sample sketch here

```
#include <TinyGPS.h>
#include <TheThingsNetwork.h>

SoftwareSerial loraSerial(2, 3);

TinyGPS gps;
SoftwareSerial ss(10, 11);

TheThingsNetwork ttn(loraSerial, debugSerial, /* TTN_FP_EU868 or TTN_FP_US915 */);
...
void setup(){
  debugSerial.begin(115200);
  loraSerial.begin(57600);
  ss.begin(9600);

 ....
}

```
Since there are 2 SoftwareSerial are being used, from code above ss RX port will be listened prior to loraSerial. So all the commands sent by loraSerial will not get response,resulting the error. Even if we switch ss and loraSerial during the setup, in later program we will need to put listen() function strategically to make sure all devices are happy and get the data from their respective RX port.

Pull request might not needed, but may be we can do something about it, like creating examples for 2 or more SoftwareSerial case? This was my test on Arduino Uno with LoRa shield + GPS shield, both shields using UART.

htdvisser:
@johanstokking?

PR #85, opened 2016-10-10
Add Serial pass through example

FokkeZB:


examples/PassThrough/PassThrough.ino
```diff
@@ -0,0 +1,26 @@
...
+  debugSerial.begin(9600);
+}
+
+void loop() {
+  if (debugSerial.available()) {
+    loraSerial.write(debugSerial.read());
+  }
+  if (loraSerial.available()) {
+    debugSerial.write(loraSerial.read());
+  }
```
johanstokking: Make it `while`; it's a bit cleaner to pass the buffers through per serial, instead of one by one if there's data available on both

FokkeZB: I'll make it `while`. Is the rest of your comment the reason why or something else I should change?

PR #86, opened 2016-10-10
sendBytes max retries

Nicolasdejean:
#62 #30

Issue #51, opened 2016-09-22
Support (max) number of retries for confirm


FokkeZB:
* [ ] Keep listening in `sendBytes()` because with confirm it will keep trying
* [ ] Use `mac set retx TTN_RETX` to set max retries (should be constant in header)

johanstokking:
Keep listening in `sendBytes()`, no 10 seconds timeout in `readLine()`.

Set `TTN_RETX` to 7 (default for RN module)

Nicolasdejean:
@johanstokking @FokkeZB Where do I use ``` mac set retx ```, in sendBytes ?

FokkeZB:
Closed by #91

PR #91, opened 2016-10-10
Sendbytes max retries

Nicolasdejean:
#86 #51

### DIFF
--- a/src/TheThingsNetwork.cpp
+++ b/src/TheThingsNetwork.cpp
@@ -174,13 +174,7 @@ int TheThingsNetwork::sendBytes(const byte* payload, int length, int port, bool 
     return -1;
   }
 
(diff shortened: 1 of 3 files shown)


FokkeZB:
Nicolas, could you rebase (not merge) on master?

src/TheThingsNetwork.cpp
```diff
@@ -178,7 +174,13 @@ int TheThingsNetwork::sendBytes(const byte* payload, int length, int port, bool
     return -1;
   }
 
-  String response = readLine();
+  String response = "";
+  if (confirm) {
+    while ((response = readLine()) == "");
```
FokkeZB: readLine no longer returns empty. Get rid of this if case and merge line 177 and 182

PR #84, opened 2016-10-10
Track airtime

Nicolasdejean:
#65 #30

PR #93, opened 2016-10-10
Track air Time

Nicolasdejean:
#84


src/TheThingsNetwork.cpp
```diff
@@ -234,6 +244,55 @@ int TheThingsNetwork::poll(int port, bool confirm) {
   return sendBytes(payload, 1, port, confirm);
 }
 
+void TheThingsNetwork::fillAirtimeInfo() {
+  this->info.sf = getInfo(readValue(F("radio get sf")));
+  this->info.ps = getInfo(readValue(F("radio get prlen")));
+  this->info.band = getInfo(readValue(F("radio get bw")));
+  this->info.header = getInfo(readValue(F("radio get crc")));
+  this->info.cr = getInfo(readValue(F("radio get cr")));
```
johanstokking: Please use proper parsing of the response, not a generic magic `getInfo` method

src/TheThingsNetwork.h
```diff
@@ -28,17 +28,32 @@ typedef unsigned long   fp_ttn_t;
...
 
+typedef struct  airtime_s
+{
+  int           sf;
+  int           de;
+  int           ps;
+  int           band;
+  int           header;
+  int           cr;
+}               airtime_t;
```
johanstokking: Spacing

src/TheThingsNetwork.cpp
```diff
@@ -245,41 +245,47 @@ int TheThingsNetwork::poll(int port, bool confirm) {
...
-  this->info.sf = getInfo(readValue(F("radio get sf")));
-  this->info.ps = getInfo(readValue(F("radio get prlen")));
-  this->info.band = getInfo(readValue(F("radio get bw")));
-  this->info.header = getInfo(readValue(F("radio get crc")));
-  this->info.cr = getInfo(readValue(F("radio get cr")));
+  getInfo(readValue(F("radio get sf")), 1);
+  getInfo(readValue(F("radio get bw")), 2);
+  getInfo(readValue(F("radio get prlen")), 3);
+  getInfo(readValue(F("radio get crc")), 4);
+  getInfo(readValue(F("radio get cr")), 5);
```
johanstokking: Get rid of `getInfo()` and just read the values here

src/TheThingsNetwork.cpp
```diff
@@ -245,47 +220,33 @@ int TheThingsNetwork::poll(int port, bool confirm) {
...
+  message = readValue(F("radio get crc"));
+  if (message == F("on")) {
+    this->info.header = 1;
   }
+  debugPrintLn(this->info.header);
+
+  message = readValue(F("radio get cr"));
+  this->info.cr = (this->info.cr + message[2] - 48);
+  
+  this->info.sf >= 11 ? this->info.de = 1 : this->info.de = 0;
```
johanstokking: Do not put assignments in the `?:`. Make it readable: `this->info.de = ... ? ... : ...;`

src/TheThingsNetwork.cpp
```diff
@@ -245,47 +220,33 @@ int TheThingsNetwork::poll(int port, bool confirm) {
...
+  
+  message = readValue(F("radio get prlen"));
+  this->info.ps = this->info.ps + message[0] - 48;
+  debugPrintLn(this->info.ps);
+  
+  message = readValue(F("radio get crc"));
+  if (message == F("on")) {
+    this->info.header = 1;
   }
+  debugPrintLn(this->info.header);
```
johanstokking: What about off, should we set `header = 0`? And why printing 0 or 1 to debug, the user doesn't know what it means

src/TheThingsNetwork.cpp
```diff
@@ -245,47 +220,33 @@ int TheThingsNetwork::poll(int port, bool confirm) {
...
-
-void TheThingsNetwork::getInfo(String message, int infoType) {
   int i;
-  if (message == "") {
-    debugPrint("Information message empty");
-    return ;
+  String message = readValue(F("radio get sf"));
+  for (i = 2; message[i] && i <= 3; i++) {
+    this->info.sf = (this->info.sf + message[i] - 48) * 10;
   }
```
johanstokking: What if we're calling this function twice; it's adding to the existing `sf`. Should the `sf` be reset before the loop?

src/TheThingsNetwork.cpp
```diff
@@ -199,8 +195,13 @@ int TheThingsNetwork::sendBytes(const byte* payload, int length, int port, bool
...
     byte downlink[64];
-    for (int i = 0, d = 0; i < downlinkLength; i++, d += 2)
-      downlink[i] = TTN_HEX_PAIR_TO_BYTE(data[d], data[d+1]);
+    byte nibble;
+    byte nextNibble;
+    for (int i = 0, d = 0; i < downlinkLength; i++, d += 2) {
+      nibble = (data[d] >= 'A') ? (data[d] - 'A' + 0x0A) : (data[d] - '0');
+      nextNibble = (data[d + 1] >= 'A') ? (data[d + 1] - 'A' + 0x0A) : (data[d + 1] - '0');
+      downlink[i] = (nibble << 4) + nextNibble;
+    }
```
johanstokking: We're going to optimize this library later, so let's not replacing optimized stuff by not optimized stuff (cc @FokkeZB)

We're going back to the define `TTN_HEX_PAIR_TO_BYTE`. It won't get much faster than that. This loop is worse as it allocates to variables. You can also inline `nibble` and `nextNibble`.

src/TheThingsNetwork.cpp
```diff
@@ -195,14 +195,8 @@ int TheThingsNetwork::sendBytes(const byte* payload, int length, int port, bool
...
-    byte nibble;
-    byte nextNibble;
-    for (int i = 0, d = 0; i < downlinkLength; i++, d += 2) {
-      nibble = (data[d] >= 'A') ? (data[d] - 'A' + 0x0A) : (data[d] - '0');
-      nextNibble = (data[d + 1] >= 'A') ? (data[d + 1] - 'A' + 0x0A) : (data[d + 1] - '0');
-      downlink[i] = (nibble << 4) + nextNibble;
-    }
-    debugPrint(F("Successful transmission. Received "));
+    for (int i = 0, d = 0; i < downlinkLength; i++, d += 2)
+      downlink[i] = TTN_HEX_PAIR_TO_BYTE(data[d], data[d+1]);debugPrint(F("Successful transmission. Received "));
```
johanstokking: Put that `debugPrint` on another line. Also use curly braces also for single line statements for readability.

Issue #97, opened 2016-10-11
Use enum for `TTN_FP_` instead of typedef


FokkeZB:
```c
#define TTN_FP_EU868 1
#define TTN_FP_US915 2
// ..
typedef unsigned long   fp_ttn_t;
// ..
fp_ttn_t fp;
```

should become

```c
enum TTN_FP: byte
{
  TTN_FP_EU868
  TTN_FP_US915
}
// ..
TTN_FP fp;
```

so that the constructor can check for valid args:

```
TheThingsNetwork(Stream& modemStream, Stream& debugStream, TTN_FP fp, int sf = TTN_DEFAULT_SF, int fsb = TTN_DEFAULT_FSB); 
```

johanstokking:
The constructor can take `fp_ttn_t` already. Also, just name it `enum ttn_fp_t` to follow naming conventions. Uppercase is reserved for constants.

Nicolasdejean:
@FokkeZB So if they are not define anymore, do I remove TTN_FP_EU868 and TTN_FP_US915 from keywords.txt ?

FokkeZB:
Closed by #101

PR #101, opened 2016-10-11
Replaced typedef and defines by enum

Nicolasdejean:
Replaced typedef and defines with enum. TTN_EU_868 has the value of 0 and  TTN_US_915 has the value of 1
#97


keywords.txt
```diff
@@ -33,5 +33,3 @@ TTN_DEFAULT_FSB         LITERAL1
 TTN_ADR_SUPPORTED       LITERAL1
 TTN_PWRIDX_868          LITERAL1
 TTN_PWRIDX_915          LITERAL1
-TTN_FP_EU868            LITERAL1
-TTN_FP_US915            LITERAL1
```
johanstokking: Aren't these being used anymore?

Nicolasdejean: @johanstokking I thought those were for the defines ?

johanstokking:
Fix conflicts too, can't merge

Issue #98, opened 2016-10-11
Move `TTN_HEX_` to cpp


FokkeZB:
`TTN_HEX_PAIR_TO_BYTE` is only used on one place in the `.cpp` and should not be defined in the header.

PR #102, opened 2016-10-12
Moved all "TTN_HEX_" to cpp file

Nicolasdejean:
Replaced all TTN_HEX_ defines to the cpp files. #98

### DIFF
--- a/src/TheThingsNetwork.cpp
+++ b/src/TheThingsNetwork.cpp
@@ -189,8 +189,13 @@ int TheThingsNetwork::sendBytes(const byte* payload, int length, int port, bool 
     String data = response.substring(portEnds + 1);
     int downlinkLength = data.length() / 2;
     byte downlink[64];
-    for (int i = 0, d = 0; i < downlinkLength; i++, d += 2)
-      downlink[i] = TTN_HEX_PAIR_TO_BYTE(data[d], data[d+1]);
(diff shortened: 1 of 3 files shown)


src/TheThingsNetwork.cpp
```diff
@@ -189,8 +189,13 @@ int TheThingsNetwork::sendBytes(const byte* payload, int length, int port, bool
...
     byte downlink[64];
-    for (int i = 0, d = 0; i < downlinkLength; i++, d += 2)
-      downlink[i] = TTN_HEX_PAIR_TO_BYTE(data[d], data[d+1]);
+    byte nibble;
+    byte nextNibble;
+    for (int i = 0, d = 0; i < downlinkLength; i++, d += 2) {
+      nibble = (data[d] >= 'A') ? (data[d] - 'A' + 0x0A) : (data[d] - '0');
+      nextNibble = (data[d + 1] >= 'A') ? (data[d + 1] - 'A' + 0x0A) : (data[d + 1] - '0');
+      downlink[i] = (nibble << 4) + nextNibble;
+    }
```
johanstokking: Optimization-wise, it's better to work with the defines as they do not allocate local variables `nibble` and `nextNibble` and are afaik readable enough to keep.

FokkeZB:
Replaces by #107

Issue #81, opened 2016-10-07
Example for sending measurements using protobufs


johanstokking:

johanstokking:
Here's a starter: https://github.com/johanstokking/ttn-protobuf-demo

FokkeZB:
Done by #108

Issue #80, opened 2016-10-07
Makefile for compiling generic protobufs to C


johanstokking:

FokkeZB:
Done by #108

Issue #79, opened 2016-10-07
Design generic Measurement protobuf


johanstokking:
E.g.:

```
message Measurement {
   optional float temperature = 100;
   optional float humidity = 101;

   repeated uint32 analog_readings = 110;
}
```

FokkeZB:
I couldn't sleep tonight and had an idea for a Message class that could contain any number of values, encoded in minimal bytes based on max range and min precision.

https://github.com/TheThingsNetwork/arduino-device-lib/wiki/Proposal:-Message-Class

johanstokking:
@FokkeZB this is what protobufs will be for 👍

This allows for a generic, reusable `Message` that contains many optional fields and arrays for custom data. The device sets the fields it needs and the protobuf marshaler only puts the used values in the buffer with a tiny header. It also supports variable length integers.

I have a demo ready shortly.

johanstokking:
@FokkeZB see some example code in #81. The file https://github.com/johanstokking/ttn-protobuf-demo/blob/master/api/measurement.proto contains `required` fields, but they will be `optional` and a long list of things we can send. We can also send variable length arrays of analog readings, but putting values in the right fields helps upstream.

FokkeZB:
wow, I did quite some C coding lately, but that puzzles me 😛

FokkeZB:
Done by #108

Issue #78, opened 2016-10-07
Integrate Nanopb in library


johanstokking:

FokkeZB:
Done by #108

PR #108, opened 2016-10-19
Protocol Buffer (nanopb) integration

Nicolasdejean:
Integrated nanopb lib, created Makefile for .proto file, remade the example, created a new class `TheThingsMessage` and new file with new fonctions for the example.  #78 #79 #80 #81


API.md
```diff
@@ -126,3 +126,53 @@ bool provision(const byte appEui[8], const appKey[16]);
...
+void encodeMessage(api_Measurement measurement);
+```
+
+- `api_Measurement measurement`: Structure with all the variable we can send.
+
+## Method: sendMessage
+Calls `sendBytes()` with the payload and the length of the payload to send a message.
+
+```c
+int sendMessage(int port = 1, bool confirm = false);
```
johanstokking: We need to start defining the numerical data types explicitly. So here it is `uint8_t` for `port`.

Also, default is 100; because `>= 100` the Handler will start decoding the message

API.md
```diff
@@ -126,3 +126,53 @@ bool provision(const byte appEui[8], const appKey[16]);
 
 - `const byte appEui[8]`: Application Identifier for the device.
 - `const byte appKey[16]`: Application Key assigned to the device.
```
johanstokking: Call it `AppEUI` and `AppKey`; these are the LoRaWAN terms.

API.md
```diff
@@ -126,3 +126,53 @@ bool provision(const byte appEui[8], const appKey[16]);
...
+
+## Class: TheThingsMessage
+
+```c
+#include <TheThingsNetwork.h>
```
johanstokking: `TheThingsMessage.h` no?

API.md
```diff
@@ -126,3 +126,53 @@ bool provision(const byte appEui[8], const appKey[16]);
...
+
+TheThingsMessage msg(Stream& modemStream);
+```
+
+- `Stream& debugStream`: Stream to write debug logs to (for The Things Node/Uno use `Serial` and data rate `9600`).
```
johanstokking: Baud rate

Also, where's the `debugStream` in the constructor? It only says `modemStream`

API.md
```diff
@@ -126,3 +126,53 @@ bool provision(const byte appEui[8], const appKey[16]);
...
+## Method: ProcessMessage
+Decode the Message received.
+
+```c
+void ProcessMessage(const byte *buffer, int size, int port);
+```
+
+- `const byte *buffer`: Bytes received.
+- `int size`: The number of bytes.
+- `int port = 1`: The port to address. Defaults to `1`.
```
johanstokking: I see no default.

Make `size` of type `size_t`

examples/ProtoBuffer/protobuffer.ino
```diff
@@ -0,0 +1,70 @@
...
+
+  msg.sendMessage();
+
+  delay(10000);
+}
+
+void message(const byte* payload, int length, int port, bool confirm) {
+  if (ttn.sendBytes(payload, length, port, confirm) == 2) {
+    Serial.println("-- MESSAGE");
+    Serial.print("Received " + String(length) + " bytes on port " + String(port) + ":");
```
johanstokking: Do not concatenate strings; call `print` for every field/text

examples/ProtoBuffer/protobuffer.ino
```diff
@@ -0,0 +1,70 @@
...
+
+  msg.encodeMessage(measurement);
+
+  msg.sendMessage();
+
+  delay(10000);
+}
+
+void message(const byte* payload, int length, int port, bool confirm) {
+  if (ttn.sendBytes(payload, length, port, confirm) == 2) {
```
johanstokking: `2` is a magic number. Should be a `#define`

src/TheThingsNetwork.h
```diff
@@ -56,4 +68,23 @@ class TheThingsNetwork
     int poll(int port = 1, bool confirm = false);
 };
 
+class TheThingsMessage
```
johanstokking: Move to `TheThingsMessage.h`

src/TheThingsNetwork.h
```diff
@@ -17,6 +21,14 @@
 #define TTN_HEX_CHAR_TO_NIBBLE(c) ((c >= 'A') ? (c - 'A' + 0x0A) : (c - '0'))
 #define TTN_HEX_PAIR_TO_BYTE(h, l) ((TTN_HEX_CHAR_TO_NIBBLE(h) << 4) + TTN_HEX_CHAR_TO_NIBBLE(l))
 
+enum ttn_message_t {
+  water,
+  motion,
+  temperature,
+  humidity,
+  all
+};
```
johanstokking: I think we can educate our users to use the `has_xxx` flag

src/TheThingsNetwork.h
```diff
@@ -56,4 +68,23 @@ class TheThingsNetwork
...
+{
+  private:
+    ttn_message_t message;
+    Stream* printStream;
+    pb_ostream_t sendStream;
+    byte buffer[51];
+    void (* messageCallback)(const byte* payload, int length, int port, bool confirm);
+
+  public:
+    TheThingsMessage(Stream& printStream);
```
johanstokking: What is `printStream`? We're using `modemStream` or `debugStream` everywhere

src/TheThingsNetwork.h
```diff
@@ -56,4 +68,23 @@ class TheThingsNetwork
...
+    void ChoseMessage(api_Measurement *measurement, ttn_message_t message);
+    void ProcessMessage(const byte *buffer, int size, int port);
```
johanstokking: Inconsistent casing

src/api/measurement.proto
```diff
@@ -0,0 +1,12 @@
+syntax = "proto2";
+
+package api;
+
+message Measurement {
+        optional uint32 motion = 101;
+	optional uint32 water = 102;
+	optional uint32 temperature = 103;
+        optional float humidity = 104;
```
johanstokking: Spacing.

Also, starts fields from 1. Temperature is a `float` as well. Make two versions; `temperature_celcius` and `temperature_fahrenheit`

src/TheThingsMessage.cpp
```diff
@@ -0,0 +1,105 @@
...
+    debugPrint(Temperature);
+    debugPrint("; ");
+  }
+  debugPrintLn();
+}
+
+void TheThingsMessage::encodeMessage(api_Measurement measurement) {
+  this->sendStream = pb_ostream_from_buffer(this->buffer, sizeof(this->buffer));
+  pb_encode(&this->sendStream, api_Measurement_fields, &measurement);
+}
```
johanstokking: Why is `encodeMessage` a separate step for the user?

src/TheThingsMessage.cpp
```diff
@@ -0,0 +1,105 @@
...
+
+void TheThingsMessage::onMessage(void (*cb)(const byte* payload, int length, int port, bool confirm)) {
+  this->messageCallback = cb;
+}
+
+int TheThingsMessage::sendMessage(int port, bool confirm) {
+  if (this->messageCallback) {
+    this->messageCallback(this->buffer, this->sendStream.bytes_written, port, confirm);
+  }
```
johanstokking: Why not send it to The Things Network here?

Nicolasdejean: That is what I wanted to do at first but we would have to bring every fonction from the class `TheThingsNetwork` to `theThingsMessage` and in the file `TheThingsMessage.cpp` like: `readValue()`, `readLine`, `sendComand()`. I thought it was better to do this than have exactly the same function in `TheThingsNetwork.cpp` and `TheThingsMessage.cpp`

johanstokking:
Also rebase

src/TheThingsMessage.cpp
Nicolasdejean: All those fonction are private in the ` TheThingsNetwork ` class and doesn't allow `TheThingsMessage` to use them, should I put them in public ?

API.md
```diff
@@ -126,3 +126,53 @@ bool provision(const byte appEui[8], const appKey[16]);
...
+```
+
+- `int port = 1`: The port to address. Defaults to `1`.
+- `bool confirm = false`: Whether to ask for confirmation. Defaults to `false`.
+
+## Method: ChoseMessage
+Calls the correct build-in message to send.
+
+```c
+void ChoseMessage(api_Measurement *measurement, ttn_message_t message);
```
FokkeZB: choseMessage

API.md
FokkeZB: The Ardunino docs say "data rate", in baud

API.md
```diff
@@ -126,3 +126,53 @@ bool provision(const byte appEui[8], const appKey[16]);
...
+## Method: encodeMessage
+Encode the message we want to send.
+
+```c
+void encodeMessage(api_Measurement measurement);
+```
+
+- `api_Measurement measurement`: Structure with all the variable we can send.
+
+## Method: sendMessage
```
FokkeZB: I would leave the sending to `TheThingsNetwork`. I'd rather add a `ttn.sendMessage(TheThingsMessage msg, uint8_t port)`, which calls `msg.getBytes()` and `msg.getLength()` to get the payload.

examples/ProtoBuffer/protobuffer.ino
```diff
@@ -0,0 +1,70 @@
...
+  pinMode(2, INPUT);
+
+  loraSerial.begin(57600);
+  debugSerial.begin(9600);
+
+  // Wait a maximum of 10s for Serial Monitor
+  while (!debugSerial && millis() < 10000);
+
+  //Set callback for incoming message
+  msg.onMessage(message);
```
FokkeZB: Leave it `ttn.onMessage()` and in the the callback let user do `TheThingsMessage msg(payload, size, port)`

src/TheThingsMessage.cpp
```diff
@@ -0,0 +1,105 @@
...
+      measurement->has_motion = 1;
+      measurement->has_humidity = 1;
+      measurement->has_temperature = 1;
+      break;
+    default:
+      return;
+  }
+}
+
+void TheThingsMessage::ProcessMessage(const byte *buffer, int size, int port) {
```
FokkeZB: processMessage, no uppercase P

src/TheThingsMessage.cpp
```diff
@@ -0,0 +1,105 @@
+#include <TheThingsNetwork.h>
+
+#define debugPrintLn(...) { if (printStream) printStream->println(__VA_ARGS__); }
+#define debugPrint(...) { if (printStream) printStream->print(__VA_ARGS__); }
+
+void TheThingsMessage::ChoseMessage(api_Measurement *measurement, ttn_message_t message) {
```
FokkeZB: choseMessage, no uppercase C

API.md
johanstokking: This requires `TheThingsNetwork` to include `TheThingsMessage`, which is beautiful but adds to the size. I'd rather have a `send()` in `TheThingsMessage` that calls `sendBytes()` of `TheThingsNetwork`

FokkeZB:
I don't like it that `TheThingsMessage` does send and receive. I prefer this API:

```c
#include <TheThingsNetwork.h>
#include <TheThingsMessage.h>

const byte appEui[8] = { 0x70, 0xB3, 0xD5, 0x7E, 0xD0, 0x00, 0x11, 0x86 };
const byte appKey[16] = { 0xEF, 0x4C, 0xAA, 0x85, 0xE1, 0xB2, 0x26, 0x53, 0x98, 0xE2, 0x3C, 0xD1, 0x31, 0x50, 0xCA, 0xDF };

#define loraSerial Serial1
#define debugSerial Serial

TheThingsNetwork ttn(loraSerial, debugSerial, TTN_FP_EU868);

void setup() {
  loraSerial.begin(57600);
  debugSerial.begin(9600);

  ttn.onMessage(message);

  ttn.join(appEui, appKey);
}

void loop() {
  float humidity = 232;
  uint32_t motion = digitalRead(2) == HIGH;
  uint32_t water = 682 - analogRead(A0);
  uint32_t temperature = 345 - analogRead(A0);

  // Where does this `api_Measurement` come from? A standard? I don't like the class name
  api_Measurement measurement = api_Measurement_init_default;

  measurement.motion = motion;
  measurement.water = water;
  measurement.temperature = temperature;
  measurement.humidity = humidity;

  // Constructor to create message from measurement
  TheThingsMessage msg(measurement, all);

  // New TTN (!) method that calls msg.getBytes() and msg.getLength() to use with ttn.sendBytes()
  ttn.sendMessage(msg, 1, true);

  delay(10000);
}

void message(const byte* payload, int length, int port) {

  // Constructor to create message from buffer
  TheThingsMessage msg(payload, length, port);

  api_Measurement measurement = msg.getMeasurement();

  // Do something with the measurement
}
```

API.md
FokkeZB: Well, then I'd just let the user use `msg.getBytes()` and `msg.getLength()` to keep it clean. The way the example is now is really confusing.

johanstokking:
The reason it's called `api_Measurement` is because it's a generated file by Nanopb. We might introduce a `typedef` in `TheThingsMessage.h` to use a nice alias. Also, we may drop the `api` package name so it simply becomes `Measurement` already.

If `TheThingsNetwork` is going to take `TheThingsMessage` as parameter type, we have to include the header and all the other Nanopb stuff as well. I wonder if Arduino optimizes this away for users who don't use `TheThingsMessage`.

Can't we have a static method like `size_t TheThingsMessage::encode(Measurement *m, byte *buffer, size_t size)` so that the user provides the buffer? Then the user can pass the buffer to `TheThingsNetwork` or anywhere else. Then, we keep the classes decoupled which makes the optimizer happy.

FokkeZB:
If the user provides the buffer he needs to know what size to make it right?

Between these two:

```c
byte payload[10]; // how do I know what size?
payload = msg.encode(m, payload, 10);
ttn.sendBytes(payload, 10);
```

```c
msg.setMeasurement(m);
ttn.sendBytes(msg.getPayload(), msg.getLength);
```

I find the second much more predictable. Msg takes care of encoding/decoding measurements to/from buffer, the rest is just as current.

FokkeZB:
On the `api_Measurement`, I'd definitely drop `api_`, but would it also make sense to chose a name less prone for type errors by non-natives? Like `Data`?

johanstokking:
The size is variable, you don't know it in advance because it will be as small as possible, that's one of the beauties of protobufs. We only know it after encoding.

What about this:

```c
SensorData data = TheThingsMessage::newSensorData();
data.has_temperature_celcius = true;
data.temperature_celcius = analogRead(A0) * ... - ...;

byte *buffer;
size_t size;
TheThingsMessage::encodeSensorData(data, &buffer, &size);
ttn.sendBytes(buffer, size);
```

That encode is declared as follows:

```c
static void encodeSensorData(SensorData data, byte **buffer, size_t *size);
```

NB 1; I wonder why we need an instance of `TheThingsMessage`; we can simply use methods defined in the header file or static methods of the class.

NB 2; Regarding naming; there's data back and forth so it's good to distinguish between uplink and downlink. We can make different protos any time. Maybe `SensorData` is more clear as it's short and the scope is then sensor readings.

NB 3; When it comes to embedded programming, there's not a lot of room to make it really, really clean.

FokkeZB:
I like that API, but find it confusing that `TheThingsMessage` isn't really a message anymore, it's more like a encoder/decoder from SensorData to buffer. What I liked about using an instance is that it would wrap the SensorData object from the user:

```c
msg TheThingsMessage;
msg.data.has_temperature_celcius = true; // why do I need this and can't just set the value?
msg.data.temperature_celcius = analogRead(A0) * ... - ...;

ttn.sendBytes(msg.getBuffer(), msg.getLength());

msg.parseBuffer(buffer, length);
debugSerial.print("Temperature: ");
debugSerial.println(msg.data.temperature_celcius);
```

johanstokking:
I think it's good to reserve the possibility to introduce other protos after `SensorData`, so I wouldn't tie `TheThingsMessage` to `SensorData`, but the name becomes confusing indeed. What about just `TheThingsMessages` as static class?

Regarding the `has_` prefix: this is how protobufs work, you have to indicate that you have this optional field. If you don't specify it, it won't be send and it won't cost payload. This is what @Nicolasdejean's `choseMessage()` came from, but I don't really like it.

src/TheThingsMessage.cpp
```diff
@@ -0,0 +1,84 @@
...
+    debugPrint("; ");
+  }
+  debugPrintLn();
+}
+
+void TheThingsMessage::encodeSensorData(sensorData *data, byte **buffer, size_t *size) {
+  byte message[51];
+  pb_ostream_t sendStream = pb_ostream_from_buffer(message, sizeof(message));
+  pb_encode(&sendStream, api_Measurement_fields, data);
+  *(buffer) = message;
```
johanstokking: Remove parentheses

src/TheThingsMessage.cpp
```diff
@@ -0,0 +1,84 @@
...
+    debugPrint("; ");
+  }
+  if (measurement.has_temperature_fahrenheit) {
+    float Temperature_fahrenheit = measurement.temperature_fahrenheit;
+    debugPrint(F("temperature fahrenheit = "));
+    debugPrint(Temperature_fahrenheit);
+    debugPrint("; ");
+  }
+  debugPrintLn();
+}
```
johanstokking: Remove function

src/TheThingsMessage.cpp
```diff
@@ -0,0 +1,84 @@
...
+    float Temperature_celcius = message.temperature_celcius;
+    debugPrint(F("temperature celcius = "));
+    debugPrintLn(Temperature_celcius);
+  }
+  if (message.has_temperature_fahrenheit) {
+    float Temperature_fahrenheit = message.temperature_fahrenheit;
+    debugPrint(F("temperature fahrenheit = "));
+    debugPrintLn(Temperature_fahrenheit);
+  }
+}
```
johanstokking: Remove function

src/TheThingsMessage.h
```diff
@@ -0,0 +1,25 @@
...
+class TheThingsMessage
+{
+  private:
+    Stream* debugStream;
+    Stream* modemStream;
+
+  public:
+    TheThingsMessage(Stream& modemStream, Stream& debugStream);
+    void processMessage(const byte *buffer, int size, int port);
+    void encodeSensorData(sensorData *measurement, byte **buffer, size_t *size);
```
johanstokking: Make static

You can remove the streams and initializer

src/api/measurement.proto
```diff
@@ -0,0 +1,13 @@
+syntax = "proto2";
+
+package api;
+
+message Measurement {
+        optional bool motion = 1;
+        optional uint32 water = 2;
+        optional float temperature_celcius = 3;
+	optional float temperature_fahrenheit = 4;
```
johanstokking: Spacing

src/TheThingsMessage.cpp
Nicolasdejean: Do I print the variables in the example ? or not at all ?

src/TheThingsMessage.cpp
Nicolasdejean: Where do I decode the message if I received one ?

johanstokking: We're not receiving yet, `SensorData` is one way

README.md
```diff
@@ -9,6 +9,11 @@ This is an [Arduino Library](https://www.arduino.cc/en/Guide/Libraries) for Ardu
 * Install the library by [Using the Library Manager](https://www.arduino.cc/en/Guide/Libraries#toc3)
 * **OR** by [Importing the .zip library](https://www.arduino.cc/en/Guide/Libraries#toc4) using either the [master](https://github.com/TheThingsNetwork/arduino-device-lib/archive/master.zip) or one of the [releases](https://github.com/TheThingsNetwork/arduino-device-lib/releases) ZIP files.
 
+To use build in messages, thanks to [Protobufs](https://en.wikipedia.org/wiki/Protocol_Buffers), in the Arduino library, you might want to install the Nanopb compiler by [downloading](http://koti.kapsi.fi/%7Ejpa/nanopb/download/) it.
+After that, in your shell, just define `$NANOPB` pointing to the Nanopb compiler's path e.g.:
+
+`export NANOPB="~/dev/bin/nanopb-0.3.6-macosx-x86"`
```
johanstokking: This is only necessary for regenerating the protobufs

Check English also

examples/ProtoBuffer/protobuffer.ino
```diff
@@ -51,31 +51,28 @@ void loop()
...
   data.temperature_fahrenheit = temperature_fahrenheit;
   data.humidity = humidity;
 
-  msg.showValues(data);
-  msg.encodeSensorData(&data, &buffer, &size);
+  //Encode the message
+  TheThingsMessage::encodeSensorData(&data, &buffer, &size);
+
+  //Send the message
   ttn.sendBytes(buffer, size);
```
johanstokking: Yep I like it. What about you @FokkeZB ?

examples/ProtoBuffer/protobuffer.ino
```diff
@@ -51,31 +51,28 @@ void loop()
...
+  debugSerial.print("Motion : ");
+  debugSerial.print(motion);
+  debugSerial.print("; Humidity : ");
+  debugSerial.print(humidity);
+  debugSerial.print("; Water : ");
+  debugSerial.print(water);
+  debugSerial.print("; Temperature celcius : ");
+  debugSerial.print(temperature_celcius);
+  debugSerial.print("; Temperature farhenheit : ");
+  debugSerial.println(temperature_fahrenheit);
```
johanstokking: Put some newlines instead of concatenating in one big string.

Also, I'm not sure if we want to print all this in the first place

src/TheThingsMessage.h
```diff
@@ -11,15 +11,8 @@ typedef api_Measurement sensorData;
...
-  private:
-    Stream* debugStream;
-    Stream* modemStream;
-
   public:
-    TheThingsMessage(Stream& modemStream, Stream& debugStream);
-    void processMessage(const byte *buffer, int size, int port);
-    void encodeSensorData(sensorData *measurement, byte **buffer, size_t *size);
-    void showValues(sensorData measurement);
+     static void encodeSensorData(sensorData *measurement, byte **buffer, size_t *size);
```
johanstokking: Spacing

Where does the type `sensorData` come from? Why call it `measurement` while you call it `data` in the cpp file?

README.md
```diff
@@ -9,10 +9,7 @@ This is an [Arduino Library](https://www.arduino.cc/en/Guide/Libraries) for Ardu
 * Install the library by [Using the Library Manager](https://www.arduino.cc/en/Guide/Libraries#toc3)
 * **OR** by [Importing the .zip library](https://www.arduino.cc/en/Guide/Libraries#toc4) using either the [master](https://github.com/TheThingsNetwork/arduino-device-lib/archive/master.zip) or one of the [releases](https://github.com/TheThingsNetwork/arduino-device-lib/releases) ZIP files.
 
-To use build in messages, thanks to [Protobufs](https://en.wikipedia.org/wiki/Protocol_Buffers), in the Arduino library, you might want to install the Nanopb compiler by [downloading](http://koti.kapsi.fi/%7Ejpa/nanopb/download/) it.
-After that, in your shell, just define `$NANOPB` pointing to the Nanopb compiler's path e.g.:
-
-`export NANOPB="~/dev/bin/nanopb-0.3.6-macosx-x86"`
+To use built-in messages in the Arduino library, you might want to [download](http://koti.kapsi.fi/%7Ejpa/nanopb/download/) the [Nanopb compiler](https://koti.kapsi.fi/jpa/nanopb/).
```
johanstokking: Again, compilation is only necessary for regenerating the protobufs. This library should be stand-alone.

Apart from this basic functionality (standardizing sensor data), we might need a guide on how to generate code from custom protos, but that is beyond the scope.

Nicolasdejean:
Is this PR still good regarding the API, or do I still have to change something in the functions ?

FokkeZB:
I've tested the Send example in the PR which now has static values. It successfully encodes and sends `08011D0000F041250000AC422D0000C242	`. How do I now verify this decoded on the backend @johanstokking?

Nicolasdejean:
@FokkeZB To verify the encoded message I sent the encoded value from the dashboard and received it on the sketch.

examples/ProtoBuffer/protobuffer.ino
```diff
@@ -0,0 +1,64 @@
...
+  // Wait a maximum of 10s for Serial Monitor
+  while (!debugSerial && millis() < 10000);
+
+  debugSerial.println("-- STATUS");
+  ttn.showStatus();
+
+  debugSerial.println("-- JOIN");
+  ttn.join(appEui, appKey);
+
+  //Chose message you want to send (send the message by writing true)
```
Nicolasdejean: I thought choosing the message you want to send had it place in the ``` setup() ``` part. But I guess we could put it in the ``` loop() ``` .

FokkeZB:
Smart :) @johanstokking do you agree we should merge #117 into this branch so that we can test it together?

Issue #104, opened 2016-10-12
Use strings for keys


FokkeZB:
```c
  if(lora.set_device_address("A88D5E97")) Serial.println("Address OK");
  if(lora.set_network_session_key("4230874028704F12BC3890D32D7A0355")) Serial.println("SESSIONKEY OK");
  if(lora.set_application_session_key("75B4C26938D4271F7903E3930CD14769")) Serial.println("APPKEY OK");
```

Nicolasdejean:
@FokkeZB Just to be sure, we use strings for key only for an easier comprehension for the user, right ? If so in ```lora.set_device_address``` or those other fonctions do we should convert the strings in a ``` const byte * ``` , am I understanding it correctly ?

johanstokking:
The snippet is an example from another library.

We want the user to provide the `appEui` as string instead of `byte appEui[8]`. There's no conversion needed, in fact the conversion can be omitted since the RN module wants the AppEUI, DevEUI and AppKey as hex string already. Just pass it as-is.

PR #110, opened 2016-10-24
Added Strings for keys

Nicolasdejean:
replaced byte * by String for the keys #104


src/TheThingsNetwork.h
```diff
@@ -48,7 +48,7 @@ class TheThingsNetwork
     String readValue(String key);
     bool sendCommand(String cmd);
     bool sendCommand(String cmd, String value);
-    bool sendCommand(String cmd, const byte* buf, int length);
+    bool sendCommand(String cmd, const byte *buf, int length);
```
FokkeZB: Uhm.. space should follow `*` right?

src/TheThingsNetwork.h
```diff
@@ -58,10 +58,10 @@ class TheThingsNetwork
     TheThingsNetwork(Stream& modemStream, Stream& debugStream, ttn_fp_t fp, int sf = TTN_DEFAULT_SF, int fsb = TTN_DEFAULT_FSB);
     void showStatus();
     void onMessage(void (*cb)(const byte* payload, int length, int port));
-    bool provision(const byte appEui[8], const byte appKey[16]);
-    bool join(const byte appEui[8], const byte appKey[16], int retries = -1, long int retryDelay = 10000);
+    bool provision(const String appEui, const String appKey);
```
FokkeZB: Tweetonig uses `const char* ` instead of `const String`... Strings are bad right?

src/TheThingsNetwork.cpp
```diff
@@ -44,11 +44,12 @@ bool TheThingsNetwork::sendCommand(String cmd) {
 }
 
 bool TheThingsNetwork::sendCommand(String cmd, String value) {
-  int l = value.length();
-  byte buf[l];
-  value.getBytes(buf, l);
+  String str = "";
```
FokkeZB: Is this change related?

examples/Workshop/Workshop.ino
```diff
@@ -6,16 +6,16 @@
 // After you registered your ABP device, go to The Things Network Dashboard
 // and copy the Device Addr, Network Session Key and App Session Key
 
-// Set your Device Address, for example: { 0x02, 0xDE, 0xAE, 0x00 };
-const byte devAddr[4] = { ... };
+// Set your Device Address, for example: "1C55B133";
+const String devAddr = " ... ";
```
FokkeZB: While you're at it, just change these constant placeholders to be like the other examples

src/TheThingsNetwork.cpp
Nicolasdejean: I figured that we are sending Strings now and we didn't need to convert them in bytes, I wasn't sure about this... Should I create another sendComand() ?

FokkeZB: Well, as it is before this PR it uses the still existing `TheThingsNetwork::sendCommand(String cmd, const byte *buf, int length)`, not this `TheThingsNetwork::sendCommand(String cmd, String value)` right?

src/TheThingsNetwork.h
johanstokking: Yep, `const char *`

src/TheThingsNetwork.cpp
```diff
@@ -44,12 +44,11 @@ bool TheThingsNetwork::sendCommand(String cmd) {
...
-  String str = "";
-  str.concat(cmd);
-  str.concat(" ");
-  str.concat(value);
+  int l = value.length();
+  byte buf[l];
+  value.getBytes(buf, l);
 
-  return sendCommand(str);
+  return sendCommand(cmd, buf, l);
```
johanstokking: We don't need to convert back to buffer

src/TheThingsNetwork.cpp
```diff
@@ -101,11 +100,18 @@ void TheThingsNetwork::onMessage(void (*cb)(const byte* payload, int length, int
...
+  String addr = "mac set devaddr ";
+  String nKey = "mac set nwkskey ";
+  String aKey = "mac set appskey ";
+
+  addr.concat(devAddr);
+  nKey.concat(nwkSKey);
+  aKey.concat(appSKey);
+  sendCommand(addr);
+  sendCommand(nKey);
+  sendCommand(aKey);
```
johanstokking: Keep the variables together, not the operations

src/TheThingsNetwork.cpp
```diff
@@ -125,9 +131,14 @@ bool TheThingsNetwork::personalize() {
...
-  sendCommand(F("mac set appeui"), appEui);
-  sendCommand(F("mac set appkey"), appKey);
+bool TheThingsNetwork::provision(const char *appEui, const char *appKey) {
+  String eui = "mac set appeui ";
+  String key = "mac set appkey ";
+
+  key.concat(appKey);
+  eui.concat(appEui);
+  sendCommand(eui);
+  sendCommand(key);
```
johanstokking: Idem

src/TheThingsNetwork.cpp
Nicolasdejean: I'm sorry but I am not sure I understand what you mean... Is it a spacing problem like that: 
```
String eui = "mac set appeui ";
String key = "mac set appkey ";
key.concat(appKey);
eui.concat(appEui);                                                                                         

sendCommand(eui);
sendCommand(key);
return sendCommand(F("mac save")); 
 ``` 
or something else ?

PR #116, opened 2016-10-26
Add init() to run once, fixes #114

### LINKED ISSUE #114: Node: LoRa module unreachable after upload
When I upload the following:

```c
#include <TheThingsNetwork.h>

const byte appEui[8] = { 0x70, 0xB3, 0xD5, 0x7E, 0xF0, 0x00, 0x00, 0xF0 };
const byte appKey[16] =  { 0x42, 0x95, 0x3C, 0x34, 0xB2, 0xAB, 0x0E, 0x03, 0x1B, 0x09, 0xDB, 0x09, 0x6C, 0x08, 0x63, 0x41 };

#define lora Serial1
#define debug Serial

TheThingsNetwork ttn(lora, debug, TTN_FP_EU868);

void setup() {  
  lora.begin(57600);
  debug.begin(9600);

  while (!debug);
  ttn.showStatus();
}

void loop()
{
  ttn.showStatus();
  
  delay(10000);
}
```

The Serial Monitor shows that the LoRa module is unreachable [here](https://github.com/TheThingsNetwork/arduino-device-lib/blob/master/src/TheThingsNetwork.cpp#L13-L20):

```
EUI: 
```

When I power off/on it works fine.

This does not seem to happen with the `Tweetonig_RN2483` library. But once the LoRa module is unreachable and I upload a sketch with the `Tweetonig_RN2483` library it also hangs. So it seems our library can get the LoRa module in a bad state.

FokkeZB:
Doing this in the constructor (which runs before `setup()`) didn't work. I can't connect to the Node anymore to do uploads or use Serial Monitor.


FokkeZB:
Travis has some issues:

- [x] _WARNING: Spurious .github folder in 'TheThingsNetwork' library_
- [ ] _warning: "BIN" redefined [enabled by default]_ https://github.com/arduino/Arduino/issues/4784_

src/TheThingsNetwork.cpp
```diff
@@ -419,7 +419,7 @@ void TheThingsNetwork::configureUS915(uint8_t sf, uint8_t fsb) {
     str = "";
     str.concat(F("mac set ch status "));
     str.concat(ch);
-    if (ch == ch500 || ch <= chHigh && ch >= chLow) {
+    if (ch == ch500 || (ch <= chHigh && ch >= chLow)) {
```
johanstokking: `&&` > `||`, keep old line

FokkeZB: It produces warnings on the CI and I agree with the CI that this is more clear

PR #117, opened 2016-10-26
Protocol Buffer(nanopb) Receive example

Nicolasdejean:
Prototype of a Receive example, added an appData.proto for downlink.


src/api/appData.proto
```diff
@@ -0,0 +1,11 @@
+syntax = "proto2";
+
+package api;
+
+message AppData {
+        optional bool motion = 1;
+	optional uint32 water = 2;
+	optional float temperature_celcius = 3;
+	optional float temperature_fahrenheit = 4;
+	optional float humidity = 5;
```
johanstokking: This is not what the app is going to send back. Please consult @FokkeZB for an example

src/TheThingsMessage.h
```diff
@@ -0,0 +1,20 @@
...
+#include "measurement.pb.h"
+#include "appData.pb.h"
+
+typedef api_Measurement sensorData;
+
+class TheThingsMessage
+{
+  public:
+    static void encodeSensorData(api_Measurement *measurement, byte **buffer, size_t *size);
+    static void processMessage(const byte *payload, size_t length, port_t port, api_Measurement *receiveData);
```
johanstokking: Follow new API as discussed with @FokkeZB

johanstokking: Also, if you typedef, use that type

src/api/measurement.proto
```diff
@@ -0,0 +1,13 @@
+syntax = "proto2";
+
+package api;
+
+message Measurement {
```
johanstokking: Rename to SensorData

src/TheThingsMessage.cpp
```diff
@@ -0,0 +1,32 @@
...
+  }
+  if (message.has_temperature_celcius) {
+    receiveData->temperature_celcius = message.temperature_celcius;
+  }
+  if (message.has_temperature_fahrenheit) {
+    receiveData->temperature_fahrenheit = message.temperature_fahrenheit;
+  }
+  if (message.has_humidity) {
+    receiveData->humidity = message.humidity;
+  }
```
johanstokking: What is this? cc @FokkeZB

FokkeZB:
We should discuss the nanobuf story with the 3 of us.

An example I've discussed with @johanstokking can be found at:
https://gist.github.com/FokkeZB/a311f9b506cffb55a1e71a32f6552294

Still some details to sort out. For me it's not clear if standard messages have a fix port or not for example.

Let's do a call tomorrow.

johanstokking:
@FokkeZB that example is good. Standard messages go on port 100 or higher. 

We can also introduce this:

```
TheThingsMessage::sendSensorData(ttn, sensorData);
```

This would encode and send it on port 100.

FokkeZB:
@Nicolasdejean please rebase on master to fix Travis:

```
git rebase master
# for every conflict, fix it and then do: git add <pathOfFixedFile> && git rebase --continue
git push --force
```

FokkeZB:
> We can also introduce this:
>
> ```c
> TheThingsMessage::sendSensorData(ttn, sensorData);
> ```
>
> This would encode and send it on port 100.

I would rather keep this separated.

Could you send sensorData over any port or does it need to be 100 for the handler to know how to decode it?

johanstokking:
@FokkeZB 100 or higher

examples/ProtoBuffer/Receive/Receive.ino
```diff
@@ -0,0 +1,91 @@
...
+    //standard message always received on port 100 or more
+    getData.appdata = api_AppData_init_default;
+    getData.has_appdata = true;
+  } else {
+    //otherwise we get a custom message
+    getData.customdata = api_CustomData_init_default;
+    getData.has_customdata = true;
+  }
+
+  TheThingsMessage::decodeGetData(&getData, payload, length);
```
johanstokking: Just like we have a `encodeSensorData`, we have a `decodeAppData`. Custom messaging requires a custom decoder and passing fields. Maybe that's beyond the scope of this example here. Keep it simple

src/TheThingsMessage.cpp
```diff
@@ -0,0 +1,37 @@
...
+  if (receiveData->has_appdata) {
+    dest_struct = &receiveData->appdata;
+  } else {
+    dest_struct = &receiveData->customdata;
+  }
+  receiveData->has_appdata = false;
+  receiveData->has_customdata = false;
+  if (!pb_decode(&stream, fields, dest_struct)) {
+    return false;
+  }
```
johanstokking: See above.

What is `api_CustomData_fields`? We don't know this in the `TheThingsMessage`.

src/api/getData.proto
```diff
@@ -0,0 +1,18 @@
...
+
+message CustomData {
+        optional uint32 foo = 3;
+        optional uint32 bar = 4;
+}
+
+message GetData {
+        optional AppData appdata = 5;
+        optional CustomData customdata = 6;
+}
```
johanstokking: Get rid of `CustomData` and `GetData`

src/api/getData.proto
```diff
@@ -0,0 +1,18 @@
+syntax = "proto2";
+
+package api;
+
+message AppData {
+        optional bool motion = 1;
+        optional uint32 water = 2;
+}
```
johanstokking: `AppData` is used for downlink; the app is not going to tell these values. These are typical downlink fields, like lights to be turned on, digitals to be put high etc

FokkeZB:
Agree that we should first design the API, then code it. Also, should we consider Capt'n proto, which seems much lighter? It's designed by the architect behind protobuf v2.

https://capnproto.org/

johanstokking:
Definitely good to consider, but the imo the packet size of Protobufs outweighs the faster computation time of Cap 'n Proto by far. Cap 'n Proto has fixed width integers and uses blanks for optional fields, while we need to send as little data as possible. There is support for compression in Cap 'n Proto, but then you're getting really close to what Protobufs already is. And Protobufs has more developer adoption.

The nice thing about the SensorData proto, is that we can put hundreds of (optional) fields in it, and only send the ones that the user uses. With Cap 'n Proto, it would be a massive structure of blanks that need to be compressed.

Nicolasdejean:
Maybe I did it wrong but I think there is also a problem with the size of the sketch for the examples... with the integration of the nanopb: the encoding and decoding fonctions we are at 99% use (~28, 500 bytes), so it doesn't let us much memory for the number of uplink and downlink variables and there is a limited option on how to interpret the downlink messages (AppData variables). Even if we optimize I don't think we can put everything in the examples... How can we solve this ?

FokkeZB:
Mm, I know bananas about Arduino/C/C++ memory efficiency. Maybe @svenhaitjema should take a look?

src/TheThingsMessage.cpp
```diff
@@ -0,0 +1,31 @@
...
+bool TheThingsMessage::decodeAppData(appdata_t *receiveData, const byte *payload, size_t length) {
+  pb_istream_t stream = pb_istream_from_buffer(payload, length);
+  if (!pb_decode(&stream, api_AppData_fields, receiveData)) {
+    return false;
+  }
+  if (receiveData->light) {
+    digitalWrite(TTN_PIN_LED, HIGH);
+  } else {
+    digitalWrite(TTN_PIN_LED, LOW);
+  }
```
johanstokking: Missing normal return statement

Also cut the light on/off

src/TheThingsMessage.cpp
```diff
@@ -0,0 +1,31 @@
...
+
+appdata_t TheThingsMessage::initAppData() {
+  appdata_t appData = api_AppData_init_default;
+  return appData;
+}
+
+sensordata_t TheThingsMessage::initSensorData() {
+  sensordata_t sensorData = api_SensorData_init_default;
+  return sensorData;
+}
```
johanstokking: Delete to reduce size

src/TheThingsMessage.cpp
```diff
@@ -0,0 +1,31 @@
+#include <TheThingsMessage.h>
+
+appdata_t TheThingsMessage::initAppData() {
+  appdata_t appData = api_AppData_init_default;
+  return appData;
+}
```
johanstokking: Delete to reduce size

src/TheThingsMessage.cpp
```diff
@@ -0,0 +1,31 @@
...
+  }
+  if (receiveData->light) {
+    digitalWrite(TTN_PIN_LED, HIGH);
+  } else {
+    digitalWrite(TTN_PIN_LED, LOW);
+  }
+}
+
+int TheThingsMessage::encodeSensorData(sensordata_t *data, byte **buffer, size_t *size) {
+  byte message[51];
```
johanstokking: Use const for size.

Maybe even better to let the user provide the buffer instead of allocating it here on the stack

src/TheThingsMessage.cpp
```diff
@@ -0,0 +1,31 @@
...
+  pb_ostream_t sendStream = pb_ostream_from_buffer(message, sizeof(message));
+  pb_encode(&sendStream, api_SensorData_fields, data);
```
johanstokking: Error handling?

src/TheThingsMessage.cpp
Nicolasdejean: Tell me if I'm wrong but I don't think we can receive a buffer[...] as argument in C, it will only use the size of the pointer not the size of the buffer. Maybe to let the user chose we could do: 
``` 
void TheThingsMessage::encodeSensorData(sensordata_t *data, byte **buffer, const size_t bufsize, size_t *size) 
```
or 
```
byte *TheThingsMessage::encodeSensorData(sensordata_t *data, size_t bufSize, size_t *size) 
```

FokkeZB:
We should continue this PR after #108 has been merged or merge this PR into #108 so we can work on it as one change.

FokkeZB:
Please rebase on master now #108 is merged

johanstokking:
Yep, please rebase and see what we can do about the program storage size. 99% is way too much.

src/TheThingsMessage.cpp
Nicolasdejean: I thought the HIGH set the led on, and LOW set the led off.

Nicolasdejean:
The 99% is due to the decode and encode functions:
  - Without the decode function (```pb_decode(&stream, api_AppData_fields, receiveData)```) we are at 77% (22,142 bytes knowing 28, 672 bytes is the max).
  - Without the encode function (``` pb_encode(&sendStream, api_SensorData_fields, data) ```) we are at 87% (25,018 bytes).
  - Without both we are at 64% (18,452 bytes).
64% is what we have in our normal examples (e.g. normal Receive example). I am not an expert but If we find a way to optimize little things inTheThingsNetwork's library we could lower the memory use to 95% max. The real problem is the nanopb library and I think it is already optimized.

Nicolasdejean:
Do we need to change something more in this PR (except for the optimization), or is it ok ?

FokkeZB:
OK with me!

docs/TheThingsMessage.md
```diff
@@ -45,17 +45,38 @@ You can also add other analog readings.
 Encode the message you want to send.
 
 ```c
-static void encodeSensorData(sensordata_t *data, byte **buffer, size_t *size);
+static void encodeSensorData(sensordata_t *data, byte **buffer, size_t bufSize, size_t *size);
```
johanstokking: If you're already passing the size of the buffer to allocate, then let the user allocate the buffer and pass the buffer size in `*size`. I think it's nice if `encodeSensorData()` does the allocation, but then don't bother the user with the allocation, so use a `const` somewhere with the to-be-allocated bytes

src/api/appData.proto
```diff
@@ -0,0 +1,7 @@
+syntax = "proto2";
+
+package api;
+
+message AppData {
+        optional bool light = 1;
+}
```
johanstokking: Line endings. Install a plugin in your editor to do this automatically

FokkeZB:
Time to do a nasty rebase on this one and close it.

Nicolasdejean:
You sure about the rebase ? for me it is up to date when I try one.

PR #130, opened 2016-11-03
Optimization

Nicolasdejean:
little optimizations in TheThingsNetwork.cpp

### DIFF
--- a/src/TheThingsNetwork.cpp
+++ b/src/TheThingsNetwork.cpp
@@ -9,6 +9,25 @@
 #define TTN_HEX_CHAR_TO_NIBBLE(c) ((c >= 'A') ? (c - 'A' + 0x0A) : (c - '0'))
 #define TTN_HEX_PAIR_TO_BYTE(h, l) ((TTN_HEX_CHAR_TO_NIBBLE(h) << 4) + TTN_HEX_CHAR_TO_NIBBLE(l))
 
+#define ON "on"
+#define OFF "off"
+#define OK "ok"
+#define ACCEPTED "accepted"
+#define MAC_TX_OK "mac_tx_ok"
+#define MAC_RX "mac_rx"
+
+#define INVALID_SF "Invalid SF"
+#define INVALID_FP "Invalid frequency plan"
+#define UNEXPECTED_RESPONSE "Unexpected response: "
+#define SEND_COMMAND_FAILED "Send command failed"
+#define JOIN_FAILED "Send join command failed"
+#define JOIN_NOT_ACCEPTED "Join not accepted: "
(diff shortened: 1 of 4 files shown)


src/TheThingsNetwork.cpp
```diff
@@ -177,10 +177,7 @@ bool TheThingsNetwork::join(int8_t retries, uint32_t retryDelay) {
   str.concat(F("mac set deveui "));
   str.concat(devEui);
   sendCommand(str);
-  while (retries != 0) {
-    if (retries > 0) {
-      retries--;
-    }
+  for (size_t i = retries; i--; ) {
```
johanstokking: You can cut even `i`:

`while (--retries >= 0)`

src/TheThingsNetwork.h
```diff
@@ -6,13 +6,130 @@
...
+const char sys_get[] PROGMEM = "get";
+const char sys_get_ver[] PROGMEM = "ver";
+const char sys_get_vdd[] PROGMEM = "vdd";
+const char sys_get_hweui[] PROGMEM = "hweui";
+const char sys_set_get_nvm[] PROGMEM = "nvm";
+const char sys_set_pindig[] PROGMEM = "pindig";
+
+const char* const sys_table[] PROGMEM = {sys_prefix,sys_sleep,sys_reset,sys_erase_fw,sys_factory_rst,sys_set,sys_get,sys_get_ver,sys_get_vdd,sys_get_hweui,sys_set_get_nvm,sys_set_pindig};
+
+#define SYS_PREFIX 0
```
FokkeZB: All these defines will be public. Can't we move this to the `cpp` file?

Nicolasdejean: Sure we can, but won't the `cpp` file be kind of messy after ?

FokkeZB: It will, but we should only make stuff public that the public needs.

FokkeZB:
Just used it and works great, with 53% storage used (instead of 95%!)

johanstokking:
@FokkeZB merge when you think it's ready

src/TheThingsNetwork.h
```diff
@@ -6,13 +6,35 @@
...
 #define TTN_DEFAULT_FSB 2
-#define TTN_RETX 7
+#define TTN_RETX "7"
 
-#define TTN_PWRIDX_868 1
-#define TTN_PWRIDX_915 5
+#define TTN_PWRIDX_868 "1"
+#define TTN_PWRIDX_915 "5"
+
+#define TTN_BUFFER_SIZE 300
```
FokkeZB: Please check if there are more defines that can be moved to `.cpp` to not be public. Only those needed to change default settings should be here, and also added to `keywords.txt`.

PR #132, opened 2016-11-10
Configure 868 order change

jpmeijers:
"mac set ch status XX on" needs to be done last. Especially after "mac set ch dcycle", otherwise the channel stays off. This commit changes the order of the configuration code to fix this.


FokkeZB:
Did you test it @Nicolasdejean? Looks ok @johanstokking?

Nicolasdejean:
Yes tested it, works fine.

PR #136, opened 2016-11-17
Fix sendPayload() and encodeSensorData()

Nicolasdejean:
fixed sendPayloads (port: int to char), found error when passing *size of the buffer in the encodeSensorData() fixed it for the moment trying to find a solution with this.


FokkeZB:
> trying to find a solution with this.

should we wait for this?

Nicolasdejean:
@FokkeZB No just merge this it fixes the library, I'll open another PR in which the user will be able to choose the size of the buffer.

PR #137, opened 2016-11-17
Waternet use case

Nicolasdejean:
I just did a PR for the review to be sure I'm going the right way... added GPS and waterLevel fields to sensorData.proto and flush field to appData.proto, created function that reads the ping ultra sound sensor and convert to cm. I was wondering should I create other .proto files just for the use Case ?


johanstokking:
This is not the right repository

Issue #139, opened 2016-11-18
Use of Serial1 in library


johanstokking:
We should always pass the stream instead of hardcoding `Serial1`.

Nicolasdejean:
closed by #140

PR #140, opened 2016-11-18
replace Serial1 by modemStream

Nicolasdejean:
#139


FokkeZB:
It's still assigned in the constructor:
https://github.com/TheThingsNetwork/arduino-device-lib/blob/6f5bea7febeaa542de5b9e4079a69874f65aded8/src/TheThingsNetwork.cpp#L190

And we've let it up to the user to do `begin()`

Issue #133, opened 2016-11-16
Provide (an enum and) defines for success/error codes


FokkeZB:
Currently `sendBytes()` [returns various integers](https://github.com/TheThingsNetwork/arduino-device-lib/blob/master/docs/TheThingsNetwork.md#method-sendbytes) as success/error codes. We should provide defines so that the user don't need to know the exact integer.

```c
#define TTN_ERROR_COMMANDFAILED -1
// etc
```

It would be nice to also have an enum with these values so that we can be clear on what the method can return.

PR #141, opened 2016-11-18
Added defines for the return values of sendBytes.

Nicolasdejean:
[DO NOT MERGE] I didn't understand what you wanted to do with an enum ? #133

### DIFF
--- a/src/TheThingsNetwork.h
+++ b/src/TheThingsNetwork.h
@@ -19,6 +19,13 @@
 
 typedef uint8_t port_t;
 
+enum ttn_response_t {
+  TTN_ERROR_SEND_COMMAND_FAILED = (-1),
+  TTN_ERROR_UNEXPECTED_RESPONSE = (-10),
+  TTN_SUCCESSFUL_TRANSMISSION = 1,
+  TTN_SUCCESSFUL_RECEIVE = 2
+};
+
 enum ttn_fp_t {
   TTN_FP_EU868,
   TTN_FP_US915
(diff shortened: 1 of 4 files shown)


FokkeZB:
Like [`ttn_fp_t`](https://github.com/TheThingsNetwork/arduino-device-lib/blob/8a96a4203c31fd016fb8737e92019e354e68de21/src/TheThingsNetwork.h#L34) this would assure the user that the function will only return variables from that type.

src/TheThingsNetwork.h
```diff
@@ -15,10 +15,22 @@
 #define TTN_PWRIDX_868 "1"
 #define TTN_PWRIDX_915 "5"
 
+#define TTN_ERROR_SEND_COMMAND_FAILED (-1)
+#define TTN_ERROR_UNEXPECTED_RESPONSE (-10)
+#define TTN_SUCCESSFUL_TRANSMISSION 1
+#define TTN_SUCCESSFUL_RECEIVE 2
+
 #define TTN_BUFFER_SIZE 300
```
johanstokking: What's this?

src/TheThingsNetwork.h
```diff
@@ -43,6 +55,7 @@ class TheThingsNetwork
     String model;
     airtime_t info;
     float airtime;
+    /* ttn_returnValue_t returnValue; */
```
johanstokking: Remove

src/TheThingsNetwork.h
```diff
@@ -15,10 +15,22 @@
...
 #define TTN_BUFFER_SIZE 300
 
 typedef uint8_t port_t;
 
+/* enum ttn_returnValues_t { */
+/*   TTN_ERROR_SEND_COMMAND_FAILED, */
+/*   TTN_ERROR_UNEXPECTED_RESPONSE, */
+/*   TTN_SUCCESSFUL_TRANSMISSION, */
+/*   TTN_SUCCESSFUL_RECEIVE */
+/* }; */
```
johanstokking: Use the `enum` instead of the defines

src/TheThingsNetwork.h
```diff
@@ -19,6 +19,13 @@
 
 typedef uint8_t port_t;
 
+enum ttn_returnValues_t {
```
johanstokking: Don't mix camelCasing and underscores. Make it `ttn_response_t`

johanstokking:
@FokkeZB please test and merge

PR #144, opened 2016-11-23
put the success/error message in flash memory

Nicolasdejean:


johanstokking:
OK, we're going to fix them first, see #145

FokkeZB:
@Nicolasdejean let's rebase and merge if all checks are OK

Issue #145, opened 2016-11-28
Fix compiler warnings


johanstokking:
See #144. Use continuous integration to find them

```
lib/src/TheThingsNetwork.cpp: In member function 'void TheThingsNetwork::reset(bool)':
lib/src/TheThingsNetwork.cpp:302:49: error: invalid conversion from 'const char*' to 'char*' [-fpermissive]
char *version = readValue(SYS_TABLE, SYS_RESET);
^
lib/src/TheThingsNetwork.cpp:303:49: error: invalid conversion from 'const char*' to 'char*' [-fpermissive]
valueToShow(SHOW_VERSION, subString(version, 7));
^
In file included from lib/src/TheThingsNetwork.cpp:4:0:
lib/src/TheThingsNetwork.h:73:10: error:   initializing argument 2 of 'void TheThingsNetwork::valueToShow(uint8_t, char*)' [-fpermissive]
void valueToShow(uint8_t index, char *value = NULL);
^
lib/src/TheThingsNetwork.cpp:304:40: error: invalid conversion from 'const char*' to 'char*' [-fpermissive]
char *model = subString(version, 0, 6);
^
lib/src/TheThingsNetwork.cpp:308:63: error: invalid conversion from 'const char*' to 'char*' [-fpermissive]
char *devEui = readValue(SYS_TABLE, SYS_TABLE, SYS_GET_HWEUI);
^
lib/src/TheThingsNetwork.cpp: In member function 'bool TheThingsNetwork::personalize()':
lib/src/TheThingsNetwork.cpp:332:29: error: invalid conversion from 'const char*' to 'char*' [-fpermissive]
char *response = readLine();
^
lib/src/TheThingsNetwork.cpp:339:112: error: invalid conversion from 'const char*' to 'char*' [-fpermissive]
Compiling .pioenvs/leonardo/lib/src/pb_common.o
stateMessage(SUCCESS_MESSAGE, SCS_PERSONALIZE_ACCEPTED, readValue(MAC_TABLE, MAC_CH_TABLE, MAC_CHANNEL_STATUS));
^
In file included from lib/src/TheThingsNetwork.cpp:4:0:
lib/src/TheThingsNetwork.h:72:10: error:   initializing argument 3 of 'void TheThingsNetwork::stateMessage(uint8_t, uint8_t, char*)' [-fpermissive]
void stateMessage(uint8_t type, uint8_t indexMsg, char *output = NULL);
^
[...]

PR #146, opened 2016-11-29
fixed compiling issues

Nicolasdejean:
will change subString function in other PR  #145


FokkeZB:
If we want we can later ignore certain linting options like this:

```yaml
env:
    - PLATFORMIO_CI_SRC=test/TheThingsNetwork.ino PLATFORMIO_BUILD_FLAGS=-Wno-parentheses
```

PR #149, opened 2016-11-30
fix subString

Nicolasdejean:


FokkeZB:
Rebased to verify it works with #148

PR #171, opened 2016-12-17
Add sleep command

xoseperez:
Sleep command with then sleeping time in milliseconds as a parameter. The module awakes after the specified time with an "ok" that it's not captured because you would probably want to do other things after setting the module to sleep...


xoseperez:
Changed and signed.

johanstokking:
Thanks!

PR #199, opened 2017-04-17
Asian frequency plans

jpmeijers:
Added the other higher Asian frequency plan, as well as the Korean frequency plan. Needs to be tested on a physical device before merging in.


johanstokking:
Please resolve conflict

jpmeijers:
Not according to the LoRaWAN Region Parameters doc. It implies that the channels and bandwidth is fixed by the regulator in Korea. We could however put a FSK channel on one of the current LoRa channels. But preferably we do not want LoRa on the same channel as FSK due to interference from LoRa to FSK. We think about making a change like this at a later stage, and rather keep to the LoRaWAN standard for now.

jpmeijers:
**TTN_FP_AS920_923**
Tested: Valid commands, ABP, OTAA, downlinks in RX1
Not tested: RX2 (not implemented in backend)

**TTN_FP_AS923_925**
Tested: Valid commands, ABP, OTAA, downlinks in RX1
Not tested: RX2 (not implemented in backend)

**TTN_FP_KR920_923**
Tested: Valid commands
Not tested: ABP, OTAA, downlinks - because https://github.com/TheThingsNetwork/gateway-conf/issues/7

Fixed a few issues that arose during testing.

PR #203, opened 2017-05-02
Serial and SoftwareSerial port comments

jpmeijers:
Added a description of the Stream parameters passed to TheThingsNetwork library. Elaborate about Software Serial and serial on other devices.

### DIFF
--- a/docs/TheThingsNetwork.md
+++ b/docs/TheThingsNetwork.md
@@ -12,8 +12,8 @@ Include and instantiate the TheThingsNetwork class. The constructor initialize t
 TheThingsNetwork ttn(Stream& modemStream, Stream& debugStream, fp_ttn_t fp, uint8_t sf = 7, uint8_t fsb = 2);
 ```
 
-- `Stream& modemStream`: Stream for the LoRa modem (for The Things Node/Uno use `Serial1` and data rate `57600`).
-- `Stream& debugStream`: Stream to write debug logs to (for The Things Node/Uno use `Serial` and data rate `9600`).
+- `Stream& modemStream`: Stream for the LoRa modem (see comments at the end of this document).
+- `Stream& debugStream`: Stream to write debug logs to (see comments at the end of this document).
 - `fp_ttn_fp fp`: The frequency plan: `TTN_FP_EU868` or `TTN_FP_US915` depending on the region you deploy in.
 - `uint8_t sf = 7`: Optional custom spreading factor. Can be `7` to `12` for `TTN_FP_EU868` and `7` to `10` for `TTN_FP_US915`. Defaults to `7`.
 - `uint8_t fsb = 2`: Optional custom frequency subband. Can be `1` to `8`. Defaults to `2` (for US915).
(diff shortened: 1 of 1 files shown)


docs/TheThingsNetwork.md
```diff
@@ -174,3 +174,68 @@ void sleep(unsigned long mseconds);
 ```
 
 - `unsigned long mseconds`: number of milliseconds to sleep.
+
+# Comments
+## Serial ports (Stream objects)
+The Stream objects (Serial ports) need to be initialized at the correct baud rates at the start of your `setup()` function. See [our examples](https://github.com/TheThingsNetwork/arduino-device-lib/blob/asian-frequency-plans/examples) for more details. For example:
```
johanstokking: Wrong branch

johanstokking: Shouldn't we really start with the definition of the serials, so explain a bit about the hardware serial(s) and potential software serials?

docs/TheThingsNetwork.md
```diff
@@ -174,3 +174,68 @@ void sleep(unsigned long mseconds);
...
+```
+void setup()
+{
+  loraSerial.begin(57600);
+  debugSerial.begin(9600);
+  ...
+}
+```
+
+### SodaqOne
```
johanstokking: SODAQ One

docs/TheThingsNetwork.md
```diff
@@ -174,3 +174,68 @@ void sleep(unsigned long mseconds);
...
+# Comments
+## Serial ports (Stream objects)
+The Stream objects (Serial ports) need to be initialized at the correct baud rates at the start of your `setup()` function. See [our examples](https://github.com/TheThingsNetwork/arduino-device-lib/blob/asian-frequency-plans/examples) for more details. For example:
+```
+  loraSerial.begin(57600);
+  debugSerial.begin(9600);
+```
+
+### TheThingsUno
+At the top of your sketch use
```
johanstokking: Or any Leonardo with RX/TX connected to the RN module

docs/TheThingsNetwork.md
```diff
@@ -174,3 +174,68 @@ void sleep(unsigned long mseconds);
...
+```
+void setup()
+{
+  loraSerial.begin(9600);
+  debugSerial.begin(9600);
+  ...
+}
+```
+
+SoftwareSerial does not operate correctly at high baud rates. We normally use it at 9600 baud. Because the RN2483 and RN2903 normally operates at 57600 baud, we need to switch it to 9600 baud so that we can communicate with it using 9600 baud. Luckily this is done automatically inside TheThingsNetwork Arduino library, so you as user do not have to worry about this.
```
johanstokking: Is the autobaud feature really strong enough to handle this?

johanstokking: Another tip is to use `AltSoftSerial`; which determines for you which RX/TX pins you can use, but it's reliable

jpmeijers: I prefer to stick to standard libraries, and therefore the SoftwareSerial library that comes with Arduino. AltSoftSerial has the side effect that it can cause interference with your own program when you use timers and interrupts. On a normal Arduino you should be able to use SoftwareSerial on all pins.

The autobaud() function is based on the version in my library. That one is quite stable. We can however test this one a bit better. I'll see if I can find an Arduino and RN2483 to test with.

johanstokking: Good to stick to standard libraries, but better not use them when you know they're not reliable in certain cases. AltSoftSerial is required when you want to use other (software) serial on The Things Uno, that's definitely not stable with standard libraries.

The autobaud is stable but I'm not sure if it really sets the RN module to any baud rate you want, or it merely syncs the serial interface timing so that it doesn't get corrupted on first transmissions.

jpmeijers:
Maybe the description of the Serial ports belongs better in https://github.com/TheThingsNetwork/docs/edit/master/_content/devices/arduino/usage.md than in the API reference. Should I move it?

I should also add a comment about a level shifter and/or voltage levels (3.3V vs 5V) for when someone builds their own device.

johanstokking:
Yes indeed

PR #204, opened 2017-05-04
Revert "Serial and SoftwareSerial port comments"

jpmeijers:
Reverts TheThingsNetwork/arduino-device-lib#203
because this belongs better in the Docs, where it was just added.


johanstokking:
Let's not revert everything; there's general cleanup that is here to stay

PR #207, opened 2017-06-07
Fix infinite join retry

jpmeijers:


johanstokking:
This changes behavior.

What's wrong with the original code, is that if `retries == 0`, it goes in infinite mode. I'd suggest introducing an `attempts` variable set to `retries + 1` and check `retries == -1 || attempts-- > 0`

jpmeijers:
In the original code any positive value will make it go into an infinite retry. The original check was just wrong.

The fix does not change the behaviour. Confirmed on a device:
`ttn.join(appEui, appKey, 2, 10000);` - Tries 2 times to join.
`ttn.join(appEui, appKey, 0, 10000);` - Does not try to join.
`ttn.join(appEui, appKey, -1, 10000);` - Infinitely tries to join.
`ttn.join(appEui, appKey, -2, 10000);` - Does not try to join.
`ttn.join(appEui, appKey);` - Infinitely tries to join.

jpmeijers:
I think I see the issue. "Try" vs "retry".

"Number of times to retry after failed or unconfirmed join. Defaults to -1 which means infinite."

So according to the documentation if the value is 0, you want it to try to join once.

jpmeijers:
Behaviour is now: 
`ttn.join(appEui, appKey, 2);` - Tries 3 times to join.
`ttn.join(appEui, appKey, 0);` - Tries 1 time to join.
`ttn.join(appEui, appKey, -1);` - Tries infinitely to join.
`ttn.join(appEui, appKey, -2);` - Tries 0 times to join.

Also did a minor version number bump.

PR #205, opened 2017-06-02
Fixing retries handling on join

VyvojRSnet:
There was an bug, that caused inifinity loop till join was successful.

PR #210, opened 2017-06-14
Feature/prevent inf loop

jpmeijers:
Prevent an infinite loop in the `readLine` function. An infinite loop might occur when the wrong Serial port is defined. Therefore print out an error in the `showStatus` function, returning after 10 seconds, rather than waiting indefinitely.

### DIFF
--- a/src/TheThingsNetwork.cpp
+++ b/src/TheThingsNetwork.cpp
@@ -438,7 +438,6 @@ void TheThingsNetwork::saveState()
   sendCommand(MAC_TABLE, MAC_SAVE, false);
   modemStream->write(SEND_MSG);
   debugPrintLn();
-  waitForOk();
 }
 
 void TheThingsNetwork::onMessage(void (*cb)(const uint8_t *payload, size_t size, port_t port))
--- a/src/TheThingsNetwork.cpp
+++ b/src/TheThingsNetwork.cpp
@@ -349,11 +349,9 @@ size_t TheThingsNetwork::readLine(char *buffer, size_t size)
 {
   size_t read = 0;
   uint8_t loops = 10;
-  //Try a maximum of 10x1000ms to read data.
-  while (read == 0 && loops>0)
+  while (read == 0 && loops > 0)
   {
     loops--;
-    //readBytesUntil will timeout after 1000ms if nothing is read.
     read = modemStream->readBytesUntil('\n', buffer, size);
   }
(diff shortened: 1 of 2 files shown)


jpmeijers:
`while (!read && loops--)` is indeed shorter, but I am not a fan of that style. I prefer to write longer code, which is more clear and less prone to bugs. In the end the compiler will anyway optimise it out.

`while (loops--)` is really difficult to read for someone with no or little C experience. I therefore agree with: https://github.com/TheThingsNetwork/arduino-device-lib/pull/205#issuecomment-306750723

But this comes down to style preference. If you want the library to follow your style I'll change it.

src/TheThingsNetwork.cpp
```diff
@@ -594,6 +596,11 @@ void TheThingsNetwork::showStatus()
 {
   readResponse(SYS_TABLE, SYS_TABLE, SYS_GET_HWEUI, buffer, sizeof(buffer));
   debugPrintIndex(SHOW_EUI, buffer);
+  if(strlen(buffer)<16)
```
johanstokking: Formatting

src/TheThingsNetwork.cpp
```diff
@@ -345,12 +345,12 @@ void TheThingsNetwork::clearReadBuffer()
...
 }
 
-size_t TheThingsNetwork::readLine(char *buffer, size_t size, uint8_t timeout)
+size_t TheThingsNetwork::readLine(char *buffer, size_t size, uint8_t attempts)
 {
   size_t read = 0;
-  while (read == 0 && timeout > 0)
+  while (read == 0 && attempts > 0)
   {
-    timeout--;
```
jpmeijers: `readBytesUntil` has a timeout of 1000ms. If you have multiple attempts to call that function, the attempts are directly proportional to the timeout. `timeout` makes more sense to me, because it is a value in seconds.

johanstokking:
This should only be used after resetting the module, to test that there is actually a module out there.

Otherwise, the RN module is stable enough to respond with something. Setting an arbitrary timeout, even user defined, is causing trouble by having stuff available on the serial that causes issues with request/response expectations.

jpmeijers:
The original reason I added this was to notify the user that communication with the RN was unsuccessful. This can be either because the wrong serial port was defined and passed to this library, or that the RN module is not responding (bad wiring, etc). In the current state the library will just hang indefinitely.

Issue #219, opened 2017-07-07
calling join() restore the module to default baud rate...


alexbn71:
Hello, 

I issue an "autobaud" to comunicate with RN2483 (at 9600) but when I call the join() function it calls at the beginning the reset() function that restore RN2483 to default baud rate, thus canceling the effect of my autobaud.

If I comment the code inside reset() that issue the reboot, all works fine.

My proposal is to leave the reset() function PUBLIC and leave the programmer the ability to force a reboot. 
While in the join() function leave only the setting of devEui.

Regards
Alessandro

alexbn71:
I'm sorry I was working on an old version of the library but the new one has a similar problem...

alexbn71:
this it works:

`void TheThingsNetwork::reset(bool adr)
{
  // Is pretty useless this test, the baud rate could be performed every time.	
  //if (!baudDetermined)
  //{
    autoBaud();
  //}
  
  // Issue a REBOOT using the right baud rate. The response will be garbage if ne 
  // communication is not done at default baud rate and should be discarded
  size_t length = readResponse(SYS_TABLE, SYS_RESET, buffer, sizeof(buffer));
  // After REBOOT set again the module to the right baud rate.
  autoBaud();
  // And now get all infos about the module 
  length = readResponse(SYS_TABLE, SYS_TABLE, SYS_GET_VER, buffer, sizeof(buffer));

  // buffer contains "RN2xx3[xx] x.x.x ...", splitting model from version
  char *model = strtok(buffer, " ");
  debugPrintIndex(SHOW_MODEL, model);
  char *version = strtok(NULL, " ");
  debugPrintIndex(SHOW_VERSION, version);
`

johanstokking:
@alexbn71 thanks for reporting. Can you file a PR?

alexbn71:
[TheThingsNetwork_Blason.zip](https://github.com/TheThingsNetwork/arduino-device-lib/files/1152446/TheThingsNetwork_Blason.zip)

The fix is inside TheThingsNetwork::reset no other changes are necessary to get it working.

PS Ignore the two minor improvments that you could find in the code outside "TheThingsNetwork::reset" function, that are not related to this topic. 

Alex

johanstokking:
Thanks. Can you file a [pull request](https://help.github.com/articles/about-pull-requests/) here on GitHub? Otherwise I'll do it, let me know

alexbn71:
Sure, sorry...

[...]

PR #223, opened 2017-07-17
Always sync baud on reset and after module reset

alexbn71:
When the MCU baud rate is different from default it is necessary to "autobaud" again after the module reset. Also, the response message after a reset is not readble if the MCU is working on a different baud rate. For this reason it it neccessary to issue another "autobaud" and a SYS_GET_VER to leave unchanged the behaviour of this function, that is reply with module informations.

### DIFF
--- a/src/TheThingsNetwork.cpp
+++ b/src/TheThingsNetwork.cpp
@@ -401,12 +401,11 @@ void TheThingsNetwork::autoBaud()
 
 void TheThingsNetwork::reset(bool adr)
 {
-  if (!baudDetermined)
-  {
-    autoBaud();
-  }
-
+  autoBaud();
   size_t length = readResponse(SYS_TABLE, SYS_RESET, buffer, sizeof(buffer));
+
+  autoBaud();
+  length = readResponse(SYS_TABLE, SYS_TABLE, SYS_GET_VER, buffer, sizeof(buffer));  
 
   // buffer contains "RN2xx3[xx] x.x.x ...", splitting model from version
   char *model = strtok(buffer, " ");
--- a/src/TheThingsNetwork.cpp
+++ b/src/TheThingsNetwork.cpp
@@ -401,10 +401,8 @@ void TheThingsNetwork::autoBaud()
 
 void TheThingsNetwork::reset(bool adr)
 {
-  size_t length;
-  
   autoBaud();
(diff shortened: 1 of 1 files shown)


johanstokking:
Thanks for reporting, we'll be reviewing this!

johanstokking:
Closes #219

src/TheThingsNetwork.cpp
```diff
@@ -401,13 +401,12 @@ void TheThingsNetwork::autoBaud()
...
-  if (!baudDetermined)
-  {
-    autoBaud();
-  }
-
+  autoBaud();
   size_t length = readResponse(SYS_TABLE, SYS_RESET, buffer, sizeof(buffer));
 
+  autoBaud();
+  length = readResponse(SYS_TABLE, SYS_TABLE, SYS_GET_VER, buffer, sizeof(buffer));
```
johanstokking: Declare `size_t length` here, and remove declaration and assignment at L405

alexbn71:
You mean this way?

johanstokking:
Well, we don't need the length of the first call. Since we're writing to `buffer` anyway, it may be clean to at least read it. Can you revert back to your first implementation? That's a bit cleaner than this one

alexbn71:
Although it is not generally used I prefer to keep the statement as a reminder that there is a return value...

src/TheThingsNetwork.cpp
jpmeijers: Is it necessary to call `sys get ver` here again, as the autobaud function already does it?

See: 
https://github.com/TheThingsNetwork/arduino-device-lib/blob/master/src/TheThingsNetwork.cpp#L392

alexbn71: I did not notice this call in the autobaud ... in my opinion it's a bit out of place to call the "sys get ver" in the autobaud, I would be more to delete it in that function.

jpmeijers: It is called in the autobaud function to make sure the autobaud was successful and that communication with the radio is good. If we remove it there we won't know if the autobaud was successful or not. Can you think of a better way to handle this?

alexbn71: What I meant was that it was an arbitrary choice that was not strictly necessary, while you could have left the post autobaud verification to the library user.

PR #224, opened 2017-07-17
Fix sleep time overflow

### LINKED ISSUE #220: sleep function - overflow
I'm working on an ATmega1284p with the RN2483A. When putting the device to sleep using the library the module does not go to sleep.
I'm using ttn.sleep(4294967295); where 4294967295 is the maximum value defined in the documentation. The result "sys sleep -1" is send to the RN2483, which results in nothing. 
If I use 2147483647 or less it works and sends "sys sleep 2147483647" and i can see the power usage go down. 
I guess somewhere there is a overflow in the library but I have not been able to spot where.

johanstokking:
Closes #220

### DIFF
--- a/src/TheThingsNetwork.cpp
+++ b/src/TheThingsNetwork.cpp
@@ -945,7 +945,7 @@ bool TheThingsNetwork::sendPayload(uint8_t mode, uint8_t port, uint8_t *payload,
 
 void TheThingsNetwork::sleep(uint32_t mseconds)
 {
-  if (mseconds < 100 || mseconds >= 4294967296)
+  if (mseconds < 100)
   {
     return;
   }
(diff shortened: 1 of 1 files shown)


src/TheThingsNetwork.cpp
```diff
@@ -945,7 +945,7 @@ bool TheThingsNetwork::sendPayload(uint8_t mode, uint8_t port, uint8_t *payload,
 
 void TheThingsNetwork::sleep(uint32_t mseconds)
 {
-  if (mseconds < 100 || mseconds >= 4294967296)
```
jpmeijers: Actually this check doesn't make sense. An unsigned 32 bit parameter can never be >=2^32. So doing this check doesn't have any effect.

johanstokking: That's why the line is red 😉

PR #226, opened 2017-07-18
Link check command feature

alexbn71:
Currently there is no method to check if a node is really still joined to TTN or if the gateway/s in range, is/are still up and running.
The "join bit" of "mac get status" response is not reliable... indeed if the only gateway in range is stopped this bit continues to return 1 (joined). Nothing strange since the rule is fire-and-forget.

A work around could then use "mac set linkchk nnn". 

_This function sets the time interval for the link check process to be triggered periodically. A <value> of ‘0’ will disable the link check process. When the time interval expires, the next application packet that will be sent to the server will include a link check MAC command. Refer to the LoRaWAN Specification V1.0 document for more information on the link check configuration._

Thanks to this command is it possible then get the number of gateways in range and the demodulation margin in dB returned on the last check.

NB: to not disobey at the TTN Fair Access Policy is in charge of those who use it to make the number of downlink stay within the limits allowed. The max value for "nnn" is 65.535 which corresponds to an interval of 18 hours about.

I've tested this method on my end node and gateway and it seems realiable, see this piece of log:

```
-- LINK CHECK ENABLED
-- GATEWAYS 1 - MARGIN 16
Sending: mac tx uncnf 1 08011D0000F041250000AC422D0000C242
Successful transmission
Sending: sys sleep 60000
-- LINK CHECK ENABLED
-- GATEWAYS 0 - MARGIN 255
Sending: mac tx uncnf 1 08011D0000F041250000AC422D0000C242
Successful transmission
Sending: sys sleep 60000
-- LINK CHECK ENABLED
-- GATEWAYS 0 - MARGIN 255
Sending: mac tx uncnf 1 08011D0000F041250000AC422D0000C242
Successful transmission
Sending: sys sleep 60000
-- LINK CHECK ENABLED
-- GATEWAYS 0 - MARGIN 255
Sending: mac tx uncnf 1 08011D0000F041250000AC422D0000C242
Successful transmission
Sending: sys sleep 60000
-- LINK CHECK ENABLED
-- GATEWAYS 0 - MARGIN 255
Sending: mac tx uncnf 1 08011D0000F041250000AC422D0000C242
Successful transmission
Sending: sys sleep 60000
-- LINK CHECK ENABLED
-- GATEWAYS 1 - MARGIN 17
Sending: mac tx uncnf 1 08011D0000F041250000AC422D0000C242
Successful transmission
Sending: sys sleep 60000
```

**Explanation: after the classic OTAA the values returned was "-- GATEWAYS 1 - MARGIN 16" then I stopped the "ttn-pkt-fwd" service and it became "-- GATEWAYS 0 - MARGIN 255"... then restarting again "ttn-pkt-fwd" returned to be "GATEWAYS 1 - MARGIN 17".**

### DIFF
--- a/src/TheThingsNetwork.cpp
+++ b/src/TheThingsNetwork.cpp
@@ -978,13 +978,6 @@ void TheThingsNetwork::linkcheck(uint16_t seconds)
   return waitForOk();  
 }
 
-uint8_t TheThingsNetwork::linkcheckMargin()
-{
-  readResponse(MAC_TABLE, MAC_GET_SET_TABLE, MAC_MRGN, buffer, sizeof(buffer));
-
-  return strtol(buffer, NULL, 10);
-}
-
 uint8_t TheThingsNetwork::linkcheckGateways()
 {
   readResponse(MAC_TABLE, MAC_GET_SET_TABLE, MAC_GWNB, buffer, sizeof(buffer));
@@ -992,3 +985,9 @@ uint8_t TheThingsNetwork::linkcheckGateways()
   return strtol(buffer, NULL, 10);
 }
 
+uint8_t TheThingsNetwork::linkcheckMargin()
+{
+  readResponse(MAC_TABLE, MAC_GET_SET_TABLE, MAC_MRGN, buffer, sizeof(buffer));
+
+  return strtol(buffer, NULL, 10);
+}
--- a/src/TheThingsNetwork.cpp
+++ b/src/TheThingsNetwork.cpp
@@ -975,7 +975,7 @@ void TheThingsNetwork::linkcheck(uint16_t seconds)
   modemStream->write(buffer);
   modemStream->write(SEND_MSG);
   debugPrintLn(buffer);
-  return waitForOk();  
+  waitForOk();  
 }
 
 uint8_t TheThingsNetwork::linkcheckGateways()
--- a/src/TheThingsNetwork.cpp
+++ b/src/TheThingsNetwork.cpp
@@ -968,7 +968,7 @@ void TheThingsNetwork::wake()
   autoBaud();
 }
 
-void TheThingsNetwork::linkcheck(uint16_t seconds)
+void TheThingsNetwork::linkCheck(uint16_t seconds)
 {
   clearReadBuffer();
   debugPrint(SENDING);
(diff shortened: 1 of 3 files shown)


jpmeijers:
Seems like a useful feature to me. Can someone in range of multiple gateways perhaps test this feature?

src/TheThingsNetwork.cpp
```diff
@@ -963,3 +967,32 @@ void TheThingsNetwork::wake()
 { 
   autoBaud();
 }
+
+void TheThingsNetwork::linkcheck(uint16_t seconds)
```
johanstokking: `linkCheck`

johanstokking: `seconds` means interval here, right?

src/TheThingsNetwork.cpp
```diff
@@ -963,3 +967,32 @@ void TheThingsNetwork::wake()
...
+  sendCommand(MAC_GET_SET_TABLE, MAC_LINKCHK, true);
+
+  sprintf(buffer, "%u", seconds);
+  modemStream->write(buffer);
+  modemStream->write(SEND_MSG);
+  debugPrintLn(buffer);
+  waitForOk();  
+}
+
+uint8_t TheThingsNetwork::linkcheckGateways()
```
johanstokking: `getLinkCheckGateways()`

src/TheThingsNetwork.cpp
```diff
@@ -963,3 +967,32 @@ void TheThingsNetwork::wake()
...
+{
+  readResponse(MAC_TABLE, MAC_GET_SET_TABLE, MAC_GWNB, buffer, sizeof(buffer));
+
+  return strtol(buffer, NULL, 10);
+}
+
+uint8_t TheThingsNetwork::linkcheckMargin()
```
johanstokking: `getLinkCheckMargin()`

src/TheThingsNetwork.cpp
```diff
@@ -963,3 +967,32 @@ void TheThingsNetwork::wake()
...
+{
+  readResponse(MAC_TABLE, MAC_GET_SET_TABLE, MAC_MRGN, buffer, sizeof(buffer));
+
+  return strtol(buffer, NULL, 10);
+}
```
johanstokking: Newline

src/TheThingsNetwork.cpp
alexbn71: Yes it is, no problem if you prefer to use the keyword interval.

johanstokking: No it's ok like this!

alexbn71:
Hi Johan, all done...

johanstokking:
Released in https://github.com/TheThingsNetwork/arduino-device-lib/releases/tag/v2.5.6

PR #231, opened 2018-03-01
Fixed LoRa module not waked from sleep by autoBaud sometimes

hallard:
I increased the break time condition setting up serial port to lower baud rate when writing 0 (break) on wake method
Bug could be reproduced by non stop shaking the node, with this code, seem fixed

May be we could also arrange autoBaud to reflect this change and do only one method.

### DIFF
--- a/src/TheThingsNetwork.cpp
+++ b/src/TheThingsNetwork.cpp
@@ -387,7 +387,7 @@ size_t TheThingsNetwork::readResponse(uint8_t prefixTable, uint8_t indexTable, u
   return readLine(buffer, size);
 }
 
-size_t TheThingsNetwork::checkComm()
+size_t TheThingsNetwork::checkModuleAvailable()
 {
     // Send sys get ver check we have an answer
     sendCommand(SYS_TABLE, 0, true, false);
@@ -411,10 +411,10 @@ void TheThingsNetwork::autoBaud()
     modemStream->write(0x55);
     modemStream->write(SEND_MSG);
     // check we can talk
-    length = checkComm();
+    length = checkModuleAvailable();
     
     // We succeded talking to the module ?
-    baudDetermined = length ? true : false;
+    baudDetermined = length > 0 ? true : false;
   }
   delay(100);
   clearReadBuffer();
@@ -440,7 +440,7 @@ void TheThingsNetwork::wake()
     pLora->write((uint8_t)0x55);
     pLora->flush();
     pLora->write(SEND_MSG);
-    if (checkComm()>0) {
+    if (checkModuleAvailable()>0) {
       baudDetermined = true;
       sleeping = false;
     }
--- a/src/TheThingsNetwork.cpp
+++ b/src/TheThingsNetwork.cpp
@@ -287,7 +287,7 @@ uint8_t receivedPort(const char *s)
   return port;
 }
 
-TheThingsNetwork::TheThingsNetwork(Stream &modemStream, Stream &debugStream, ttn_fp_t fp, uint8_t sf, uint8_t fsb)
+TheThingsNetwork::TheThingsNetwork(SerialType &modemStream, Stream &debugStream, ttn_fp_t fp, uint8_t sf, uint8_t fsb)
 {
   this->debugStream = &debugStream;
   this->modemStream = &modemStream;
@@ -419,23 +419,21 @@ void TheThingsNetwork::autoBaud()
 
 void TheThingsNetwork::wake()
 {  
-  HardwareSerial *pLora = (HardwareSerial*) modemStream;
-  
   // If sleeping
   if (sleeping) 
   {
     // Send a 0 at lower speed to be sure always received
     // as a character a 57600 baud rate
-    pLora->flush();
-    pLora->begin(2400);
-    pLora->write((uint8_t) 0x00);
-    pLora->flush();
+    modemStream->flush();
+    modemStream->begin(2400);
+    modemStream->write((uint8_t) 0x00);
+    modemStream->flush();
     delay(20);
     // set baudrate back to normal and send autobaud
-    pLora->begin(57600);
-    pLora->write((uint8_t)0x55);
-    pLora->flush();
-    pLora->write(SEND_MSG);
+    modemStream->begin(57600);
+    modemStream->write((uint8_t)0x55);
+    modemStream->flush();
+    modemStream->write(SEND_MSG);
     if (checkModuleAvailable() > 0) {
       baudDetermined = true;
       sleeping = false;
--- a/src/TheThingsNetwork.cpp
+++ b/src/TheThingsNetwork.cpp
@@ -409,8 +409,8 @@ void TheThingsNetwork::autoBaud()
     // check we can talk
     length = checkModuleAvailable();
     
-    // We succeded talking to the module ?
-    baudDetermined = length > 0 ;
+    // We succeeded talking to the module ?
+    baudDetermined = (length > 0) ;
   }
   delay(100);
   clearReadBuffer();
@@ -419,13 +419,13 @@ void TheThingsNetwork::autoBaud()
 
 bool TheThingsNetwork::isSleeping()
 {  
-  return baudDetermined ? false : true ;
+  return !baudDetermined;
 }
 
 void TheThingsNetwork::wake()
 {  
   // If sleeping
-  if (!baudDetermined) 
+  if (isSleeping()) 
   {
     // Send a 0 at lower speed to be sure always received
     // as a character a 57600 baud rate
@@ -650,7 +650,7 @@ void TheThingsNetwork::showStatus()
   debugPrintIndex(SHOW_RX_DELAY_2, buffer);
 }
 
-// Puting this common fonction saves 238 bytes of flash
+// Puting this common function saves 238 bytes of flash
 void TheThingsNetwork::configureChannelsFreq(uint32_t freq, uint8_t first, uint8_t last, uint8_t first_dr)
 {
   uint8_t ch;
--- a/src/TheThingsNetwork.cpp
+++ b/src/TheThingsNetwork.cpp
@@ -436,7 +436,7 @@ void TheThingsNetwork::wake()
     pLora->write((uint8_t)0x55);
     pLora->flush();
     pLora->write(SEND_MSG);
-    if (checkModuleAvailable()>0) {
+    if (checkModuleAvailable() > 0) {
       baudDetermined = true;
       sleeping = false;
     }
--- a/src/TheThingsNetwork.cpp
+++ b/src/TheThingsNetwork.cpp
@@ -430,12 +430,16 @@ void TheThingsNetwork::wake()
     // Send a 0 at lower speed to be sure always received
     // as a character a 57600 baud rate
     modemStream->flush();
+#ifdef HARDWARE_UART
     modemStream->begin(2400);
+#endif
     modemStream->write((uint8_t) 0x00);
     modemStream->flush();
     delay(20);
     // set baudrate back to normal and send autobaud
+#ifdef HARDWARE_UART
     modemStream->begin(57600);
+#endif
     modemStream->write((uint8_t)0x55);
     modemStream->flush();
     modemStream->write(SEND_MSG);
@@ -646,16 +650,16 @@ void TheThingsNetwork::showStatus()
   debugPrintIndex(SHOW_RX_DELAY_2, buffer);
 }
 
-// Puting this common fonction save 238 bytes of flash
-void TheThingsNetwork::configureChannelsFreq(uint32_t freq, uint8_t begin, uint8_t end, uint8_t start)
+// Puting this common fonction saves 238 bytes of flash
+void TheThingsNetwork::configureChannelsFreq(uint32_t freq, uint8_t first, uint8_t last, uint8_t first_dr)
 {
   uint8_t ch;
   char buf[10];
 
-  for (ch = begin; ch < end; ch++)
+  for (ch = first; ch <= last; ch++)
   {
     sendChSet(MAC_CHANNEL_DCYCLE, ch, "799");
-    if (ch > start)
+    if (ch > first_dr)
     {
       sprintf(buf, "%lu", freq);
       sendChSet(MAC_CHANNEL_FREQ, ch, buf);
@@ -704,7 +708,6 @@ void TheThingsNetwork::configureAS920_923()
    * CH0 = 923.2MHz
    * CH1 = 923.4MHz
    */
-  sendMacSet(MAC_ADR, "off"); // TODO: remove when ADR is implemented for this plan
   sendMacSet(MAC_RX2, "2 923200000");
 
   configureChannelsFreq(922000000, 0, 8, 1);
--- a/src/TheThingsNetwork.cpp
+++ b/src/TheThingsNetwork.cpp
@@ -387,6 +387,16 @@ size_t TheThingsNetwork::readResponse(uint8_t prefixTable, uint8_t indexTable, u
   return readLine(buffer, size);
 }
 
+size_t TheThingsNetwork::checkComm()
+{
+    // Send sys get ver check we have an answer
+    sendCommand(SYS_TABLE, 0, true, false);
+    sendCommand(SYS_TABLE, SYS_GET, true, false);
+    sendCommand(SYS_TABLE, SYS_GET_VER, false, false);
+    modemStream->write(SEND_MSG);
+    return modemStream->readBytesUntil('\n', buffer, sizeof(buffer));
+}
+
 void TheThingsNetwork::autoBaud()
 {
   // Courtesy of @jpmeijers
@@ -396,19 +406,45 @@ void TheThingsNetwork::autoBaud()
   while (attempts-- && length == 0)
   {
     delay(100);
+    // Send break + Autobaud 
     modemStream->write((byte)0x00);
     modemStream->write(0x55);
     modemStream->write(SEND_MSG);
-    sendCommand(SYS_TABLE, 0, true, false);
-    sendCommand(SYS_TABLE, SYS_GET, true, false);
-    sendCommand(SYS_TABLE, SYS_GET_VER, false, false);
-    modemStream->write(SEND_MSG);
-    length = modemStream->readBytesUntil('\n', buffer, sizeof(buffer));
+    // check we can talk
+    length = checkComm();
+    
+    // We succeded talking to the module ?
+    baudDetermined = length ? true : false;
   }
   delay(100);
   clearReadBuffer();
   modemStream->setTimeout(10000);
-  baudDetermined = true;
+}
+
+void TheThingsNetwork::wake()
+{  
+  HardwareSerial *pLora = (HardwareSerial*) modemStream;
+  
+  // If sleeping
+  if (sleeping) 
+  {
+    // Send a 0 at lower speed to be sure always received
+    // as a character a 57600 baud rate
+    pLora->flush();
+    pLora->begin(2400);
+    pLora->write((uint8_t) 0x00);
+    pLora->flush();
+    delay(20);
+    // set baudrate back to normal and send autobaud
+    pLora->begin(57600);
+    pLora->write((uint8_t)0x55);
+    pLora->flush();
+    pLora->write(SEND_MSG);
+    if (checkComm()>0) {
+      baudDetermined = true;
+      sleeping = false;
+    }
+  }
 }
 
 void TheThingsNetwork::reset(bool adr)
@@ -969,11 +1005,10 @@ void TheThingsNetwork::sleep(uint32_t mseconds)
   modemStream->write(buffer);
   modemStream->write(SEND_MSG);
   debugPrintLn(buffer);
-}
 
-void TheThingsNetwork::wake()
-{ 
-  autoBaud();
+  // to be determined back on wake up
+  baudDetermined = false;
+  sleeping = true;
 }
 
 void TheThingsNetwork::linkCheck(uint16_t seconds)
--- a/src/TheThingsNetwork.cpp
+++ b/src/TheThingsNetwork.cpp
@@ -410,7 +410,7 @@ void TheThingsNetwork::autoBaud()
     length = checkModuleAvailable();
     
     // We succeded talking to the module ?
-    baudDetermined = length > 0 ? true : false;
+    baudDetermined = length > 0 ;
   }
   delay(100);
   clearReadBuffer();
--- a/src/TheThingsNetwork.cpp
+++ b/src/TheThingsNetwork.cpp
@@ -390,11 +390,7 @@ size_t TheThingsNetwork::readResponse(uint8_t prefixTable, uint8_t indexTable, u
 size_t TheThingsNetwork::checkModuleAvailable()
 {
     // Send sys get ver check we have an answer
-    sendCommand(SYS_TABLE, 0, true, false);
-    sendCommand(SYS_TABLE, SYS_GET, true, false);
-    sendCommand(SYS_TABLE, SYS_GET_VER, false, false);
-    modemStream->write(SEND_MSG);
-    return modemStream->readBytesUntil('\n', buffer, sizeof(buffer));
+    return readResponse(SYS_TABLE, SYS_TABLE, SYS_GET_VER, buffer, sizeof(buffer));  
 }
 
 void TheThingsNetwork::autoBaud()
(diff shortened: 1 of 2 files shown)


src/TheThingsNetwork.cpp
```diff
@@ -396,19 +406,45 @@ void TheThingsNetwork::autoBaud()
...
-    sendCommand(SYS_TABLE, 0, true, false);
-    sendCommand(SYS_TABLE, SYS_GET, true, false);
-    sendCommand(SYS_TABLE, SYS_GET_VER, false, false);
-    modemStream->write(SEND_MSG);
-    length = modemStream->readBytesUntil('\n', buffer, sizeof(buffer));
+    // check we can talk
+    length = checkComm();
+    
+    // We succeded talking to the module ?
+    baudDetermined = length ? true : false;
```
johanstokking: `length > 0` ?

src/TheThingsNetwork.cpp
```diff
@@ -387,6 +387,16 @@ size_t TheThingsNetwork::readResponse(uint8_t prefixTable, uint8_t indexTable, u
   return readLine(buffer, size);
 }
 
+size_t TheThingsNetwork::checkComm()
+{
+    // Send sys get ver check we have an answer
+    sendCommand(SYS_TABLE, 0, true, false);
+    sendCommand(SYS_TABLE, SYS_GET, true, false);
+    sendCommand(SYS_TABLE, SYS_GET_VER, false, false);
```
johanstokking: Combine in one command

src/TheThingsNetwork.cpp
```diff
@@ -396,19 +406,45 @@ void TheThingsNetwork::autoBaud()
...
   }
   delay(100);
   clearReadBuffer();
   modemStream->setTimeout(10000);
-  baudDetermined = true;
+}
+
+void TheThingsNetwork::wake()
+{  
+  HardwareSerial *pLora = (HardwareSerial*) modemStream;
```
johanstokking: We can't do this; it's not guaranteed that `modemStream` is a `HardwareSerial`. In fact, we use quite some alternatives

src/TheThingsNetwork.h
```diff
@@ -60,6 +61,7 @@ class TheThingsNetwork
   void debugPrintIndex(uint8_t index, const char *value = NULL);
   void debugPrintMessage(uint8_t type, uint8_t index, const char *value = NULL);
 
+  size_t checkComm();
```
johanstokking: `checkModuleAvailable()`

src/TheThingsNetwork.cpp
hallard: done

src/TheThingsNetwork.cpp
hallard: done

src/TheThingsNetwork.h
hallard: done

src/TheThingsNetwork.cpp
hallard: The library is called TTN Arduino device lib, and TTN devices (The Things Node and The Things Uno) are for now connected and using hardware serial of the 32U4.
I think focus should be fixing Low Power issues (and this fix works) until we have another device using another method than hardware Serial. In this case we could resolve with some `#define` and boards.txt definition)
Anyway I'm opened to any other alternatives. Just let me know

src/TheThingsNetwork.cpp
```diff
@@ -411,10 +411,10 @@ void TheThingsNetwork::autoBaud()
     modemStream->write(0x55);
     modemStream->write(SEND_MSG);
     // check we can talk
-    length = checkComm();
+    length = checkModuleAvailable();
     
     // We succeded talking to the module ?
-    baudDetermined = length ? true : false;
+    baudDetermined = length > 0 ? true : false;
```
johanstokking: You can remove the `?:` thing now 😁

src/TheThingsNetwork.cpp
```diff
@@ -440,7 +440,7 @@ void TheThingsNetwork::wake()
     pLora->write((uint8_t)0x55);
     pLora->flush();
     pLora->write(SEND_MSG);
-    if (checkComm()>0) {
+    if (checkModuleAvailable()>0) {
```
johanstokking: Spacing; `) > 0`

src/TheThingsNetwork.cpp
johanstokking: This library is intended to use with any Microchip RN2xx3 module, whether or not on The Things Uno or The Things Node, for example SODAQ boards or Arduinos with the RN module on a shield. Therefore, we cannot introduce a dependency on specific serials, let alone a cast like this, unfortunately.

hallard: Ok, I've implemented the same HAL than in Sodaq_RN2483 library so it use 

```cpp
#if defined(ARDUINO_ARCH_AVR)
typedef HardwareSerial SerialType;
#elif defined(ARDUINO_ARCH_SAM) || defined(ARDUINO_ARCH_SAMD)
typedef Uart SerialType;
#else
typedef Stream SerialType;
#endif
```
and of course changing Stream to SerialType in the library, should do the trick.

src/TheThingsNetwork.h
```diff
@@ -20,6 +20,14 @@
 
 #define TTN_BUFFER_SIZE 300
 
+#if defined(ARDUINO_ARCH_AVR)
+typedef HardwareSerial SerialType;
+#elif defined(ARDUINO_ARCH_SAM) || defined(ARDUINO_ARCH_SAMD)
+typedef Uart SerialType;
+#else
+typedef Stream SerialType;
+#endif
```
johanstokking: The code breaks on `Stream`, because there's no `begin()`, right?

We can't introduce breaking changes in this repository.

We also have users using `AltSoftSerial`, deriving from `Stream()`, introducing `begin()`.

hallard: No in this case because Serial Type is depending on target and all defined target have `begin()`

For AltSoftSerial, it's another thing, I need to think on how to avoid problem with this one.

johanstokking: Where's `begin()` in `Stream` defined?

Not here: https://github.com/arduino/Arduino/blob/master/hardware/arduino/avr/cores/arduino/Stream.h, nor in https://github.com/arduino/Arduino/blob/master/hardware/arduino/avr/cores/arduino/Print.h

hallard: Sure the original TTN code breaks on begin() because it's a Stream. I'm just saying that this PR does not break because there is a `typedef` of `SerialType` depending on target (AVR or SAMD) using this in the include file
```
#if defined(ARDUINO_ARCH_AVR)
typedef HardwareSerial SerialType;
#elif defined(ARDUINO_ARCH_SAM) || defined(ARDUINO_ARCH_SAMD)
typedef Uart SerialType;
#else
typedef Stream SerialType;
#endif
```

and new TTN object instanceuse SerialType
```cpp
TheThingsNetwork(SerialType &modemStream, Stream &debugStream, ttn_fp_t fp, uint8_t sf = TTN_DEFAULT_SF, uint8_t fsb = TTN_DEFAULT_FSB);
```
I need to see what we can do for AltSoftSerial, but I do not see any other solution than a define.

src/TheThingsNetwork.h
```diff
@@ -8,6 +8,9 @@
 #include <Stream.h>
 #include <avr/pgmspace.h>
 
+// If you need to use soft serial library you need to uncomment this line
+//#define USE_SOFT_SERIAL
```
johanstokking: We can't expect people to uncomment lines somewhere in libraries. Also, that will be overwritten when updating them.

hallard: The problem is that we can't do that in sketch, not that I do not want, it won't just work because when the compiler compile the library it does not know anything about what is done in the sketch, nor defined values with `#define`

So a decision is on the go, using stream instead of HardwareSerial makes sense to have majority devices compatible (which are not TTN one) without changing a line of code in the library. But on the other hand it impact the LowPower mode for TTN devices that use all Hardware Serial.

From my point of view TTN lib should match TTN devices and make them working as expected (ie no impact on battery life), and I think it’s what TTN backers and devices buyers want but I may be wrong.

Another vision is to make others boards users happy and be able to use soft serial on other devices (not TTN one) without changing any line of code in the library. This will impact severely battery life for users in the case above that wait for a battery life for one year.

But it's not my decision.

Another way could be having dedicated TTN Boards (As sparkfun, Adafruit, ..) in Arduino IDE, for example `The Things Node` and 'The Things Uno`  and select these board to build and upload. This way, library know the target boards and can do some conditional compilation.

hallard: Ok, we now have a fix, Samuel created TTN boards and we tested together.

Need to install new TTN boards from Arduino IDE Boards URL. In IDE preference add this additional board manager URL
https://raw.githubusercontent.com/samuel-puschacher/TTP_Arduino_Boards/master/package_thethingsproducts_index.json 

Once added, install from board manager install TheThingsNetwork boards
image

Once done Select the correct board as target
image

If selected The Things Uno or The Things Node, it will use HardwareSerial and LowPowerFix will be active, else, stay as usual Stream.

Hope this is satisfying ?

samuel-puschacher: So I hope, we can bring this further now

samuel-puschacher: This solution is even downwards compatible ....
And the best the docs are already updated, see PR: https://github.com/TheThingsNetwork/docs/pull/226

src/TheThingsNetwork.cpp
```diff
@@ -612,18 +646,16 @@ void TheThingsNetwork::showStatus()
   debugPrintIndex(SHOW_RX_DELAY_2, buffer);
 }
 
-void TheThingsNetwork::configureEU868()
+// Puting this common fonction save 238 bytes of flash
```
johanstokking: `function saves`

src/TheThingsNetwork.cpp
```diff
@@ -668,21 +707,7 @@ void TheThingsNetwork::configureAS920_923()
   sendMacSet(MAC_ADR, "off"); // TODO: remove when ADR is implemented for this plan
```
johanstokking: Reminds me; ADR is implemented

src/TheThingsNetwork.cpp
```diff
@@ -612,18 +646,16 @@ void TheThingsNetwork::showStatus()
   debugPrintIndex(SHOW_RX_DELAY_2, buffer);
 }
 
-void TheThingsNetwork::configureEU868()
+// Puting this common fonction save 238 bytes of flash
+void TheThingsNetwork::configureChannelsFreq(uint32_t freq, uint8_t begin, uint8_t end, uint8_t start)
```
johanstokking: - Seems to make more sense to make `end` inclusive
- Unclear what `begin` vs `start` means

src/TheThingsNetwork.h
johanstokking: With in the sketch, I mean in the user sketch, the baud rate thing. But I think a board is a very good suggestion.

src/TheThingsNetwork.h
```diff
@@ -20,6 +23,19 @@
 
 #define TTN_BUFFER_SIZE 300
 
+// The Things Industries devices
```
johanstokking: The Things Products

src/TheThingsNetwork.cpp
```diff
@@ -396,19 +402,47 @@ void TheThingsNetwork::autoBaud()
...
+  {
+    // Send a 0 at lower speed to be sure always received
+    // as a character a 57600 baud rate
+    modemStream->flush();
+    modemStream->begin(2400);
+    modemStream->write((uint8_t) 0x00);
+    modemStream->flush();
+    delay(20);
+    // set baudrate back to normal and send autobaud
+    modemStream->begin(57600);
```
johanstokking: Again, `Stream` does not have `begin()`, right? If so, where is it defined?

I.e. this code is still breaking when `SerialType` cannot be determined and falls back to `Stream`.

johanstokking:
Can you file a PR in https://github.com/TheThingsProducts/arduino for the board file @samuel-puschacher ?

src/TheThingsNetwork.cpp
hallard: done

src/TheThingsNetwork.cpp
hallard: ok, so removed the line

src/TheThingsNetwork.cpp
hallard: changed name to first (first channel), last (last channel) and firs_dr (first channel to change dr)

src/TheThingsNetwork.h
hallard: done

src/TheThingsNetwork.cpp
hallard: Correct tested with a defined symbol when a hardware serial is used 
take care, now this will compile for SparkFun Micro but won't work since the begin will not be executed. For this to work users need to select new TTN boards in IDE ;-)

hallard:
@Johan,
Why you don't just Fork Samuel board repo since he's the creator and the maintainer?
I let you see with him how to organize this because it's not on any of my repo.

src/TheThingsNetwork.cpp
johanstokking: Why would that not work? In those cases, `HARDWARE_UART` gets defined on https://github.com/TheThingsNetwork/arduino-device-lib/pull/231/files/bab08ab865752ff5208c71f6f9dae807e6be8fba..a164a60fd13751bf79ae74f4b50bb9afcabd29ed#diff-49a17fc31f5e939a36dc3a3e67227c1bR30

src/TheThingsNetwork.h
```diff
@@ -23,15 +20,18 @@
...
+// The Things Products devices
 #if defined(ARDUINO_THINGS_NODE) || defined (ARDUINO_THINGS_UNO)
 typedef HardwareSerial SerialType;
+#define HARDWARE_UART
 
 #elif defined(ARDUINO_ARCH_SAM) || defined(ARDUINO_ARCH_SAMD)
 typedef HardwareSerial SerialType;
+#define HARDWARE_UART
 
 #elif defined(ARDUINO_ARCH_SAM) || defined(ARDUINO_ARCH_SAMD)
```
johanstokking: Why are you checking exactly the same thing as on L28?

johanstokking:
@hallard @samuel-puschacher I'm fine with forking if the repository is named `arduino` or `arduino-boards`

src/TheThingsNetwork.h
hallard: Because it's a mistake, it's fixed

src/TheThingsNetwork.cpp
```diff
@@ -396,19 +402,51 @@ void TheThingsNetwork::autoBaud()
...
   }
   delay(100);
   clearReadBuffer();
   modemStream->setTimeout(10000);
-  baudDetermined = true;
+}
+
+bool TheThingsNetwork::isSleeping()
+{  
+  return baudDetermined ? false : true ;
```
johanstokking: `return !baudDetermined;`

johanstokking: Also is this a sensible check? Semantically it's not clear to me.

src/TheThingsNetwork.cpp
```diff
@@ -396,19 +402,51 @@ void TheThingsNetwork::autoBaud()
...
     modemStream->write(SEND_MSG);
-    sendCommand(SYS_TABLE, 0, true, false);
-    sendCommand(SYS_TABLE, SYS_GET, true, false);
-    sendCommand(SYS_TABLE, SYS_GET_VER, false, false);
-    modemStream->write(SEND_MSG);
-    length = modemStream->readBytesUntil('\n', buffer, sizeof(buffer));
+    // check we can talk
+    length = checkModuleAvailable();
+    
+    // We succeded talking to the module ?
```
johanstokking: `succeEded`

src/TheThingsNetwork.cpp
```diff
@@ -396,19 +402,51 @@ void TheThingsNetwork::autoBaud()
...
+    baudDetermined = length > 0 ;
```
johanstokking: If a number is non-zero, it will implicitly be true when assigning to a boolean.

src/TheThingsNetwork.cpp
```diff
@@ -396,19 +402,51 @@ void TheThingsNetwork::autoBaud()
...
+
+bool TheThingsNetwork::isSleeping()
+{  
+  return baudDetermined ? false : true ;
+}
+
+void TheThingsNetwork::wake()
+{  
+  // If sleeping
+  if (!baudDetermined)
```
johanstokking: Why not call the `isSleeping()` method? It's a call, but it's semantically clearer.

src/TheThingsNetwork.cpp
```diff
@@ -612,18 +650,16 @@ void TheThingsNetwork::showStatus()
   debugPrintIndex(SHOW_RX_DELAY_2, buffer);
 }
 
-void TheThingsNetwork::configureEU868()
+// Puting this common fonction saves 238 bytes of flash
```
johanstokking: `fUnction`

src/TheThingsNetwork.cpp
```diff
@@ -612,18 +650,16 @@ void TheThingsNetwork::showStatus()
...
+void TheThingsNetwork::configureChannelsFreq(uint32_t freq, uint8_t first, uint8_t last, uint8_t first_dr)
 {
-  sendMacSet(MAC_RX2, "3 869525000");
-  sendChSet(MAC_CHANNEL_DRRANGE, 1, "0 6");
-
-  char buf[10];
-  uint32_t freq = 867100000;
   uint8_t ch;
-  for (ch = 0; ch < 8; ch++)
+  char buf[10];
```
johanstokking: Although I know that `sprintf()` doesn't allocate memory, all these small buffers on the stack do allocate extra memory. So yeah the code is smaller, and we'll leave it this way, but do note that the end result is that this code requires more (dynamic) memory.

src/TheThingsNetwork.cpp
hallard: yes it's the same thing,  changed it

src/TheThingsNetwork.cpp
hallard: done

src/TheThingsNetwork.cpp
hallard: yes because it's a test operation so it return true or false, I added some parenthesis

src/TheThingsNetwork.cpp
hallard: Correct

src/TheThingsNetwork.cpp
hallard: done

src/TheThingsNetwork.h
```diff
@@ -20,6 +20,19 @@
...
+#define HARDWARE_UART
+
+#elif defined(ARDUINO_ARCH_SAM) || defined(ARDUINO_ARCH_SAMD)
+typedef HardwareSerial SerialType;
+#define HARDWARE_UART
+
+#else
+typedef Stream SerialType;
+#endif
+
```
jpmeijers: I don't think we should do this in the library. With this code we are not allowing people to use strange setups like a second RN2483 on a SoftwareSerial port on a Things Uno.

And why are we not allowing SoftwareSerial on SAM and SAMD devices?

src/TheThingsNetwork.cpp
```diff
@@ -396,19 +402,51 @@ void TheThingsNetwork::autoBaud()
...
+  // If sleeping
+  if (isSleeping()) 
+  {
+    // Send a 0 at lower speed to be sure always received
+    // as a character a 57600 baud rate
+    modemStream->flush();
+#ifdef HARDWARE_UART
+    modemStream->begin(2400);
+#endif
+    modemStream->write((uint8_t) 0x00);
```
jpmeijers: Is the reason for all these (breaking) changes just so that we can send the break condition at a lower baudrate? Won't it be simpler to do it differently like bitbanging the break condition on the correct pins, using digitalWrite() and delay()?

src/TheThingsNetwork.cpp
jpmeijers: I really don't like using `sprintf()` on embedded devices, unless it is really necessary. I also prefer to use statically allocated memory, rather than many dynamically allocated variables. Using statically allocated ones you can know at compile time that you will run out of memory. Doing it dynamically will just cause strange behaviour down the line.

hallard: Guys,
just a a reminder, sprintf stuff has nothing to do with my PR and was present is the code before I made any changes. It was present and used, so I used it also for the same things it was already used and thought it was consistent even if I agree about dynamically allocated variables.
Charles

samuel-puschacher: Sprintf  wasnt implemented in this PR, so that would be stuff for an issue @jpmeijers

src/TheThingsNetwork.h
samuel-puschacher: We could also disable it on the Uno... but power usage will increase 
But I can't see a reason why to put a second Rn2483 on the Things Uno

src/TheThingsNetwork.cpp
hallard: The reasons of the changes are Low Power. Waking 32U4 (or other) from sleep by the watchdog every 8s for doing nothing is a non sense when you have a module such like the RN2483 than can wake you after a given time with sleep command. The wake up occur when the module sends data to the serial line (that wake up the Atmel device). Unfortunately sometimes with this technique, and we can reproduce the problem, the existing wake (autobaud) was not able to wake the module and then lock the RN2483/Atmel communication. The fix I've made with `begin()` solve this issue.
I already thought about bitbanging but touching/changing I/O port that is configured before by hardware UART stack (by 1st begin)  is not really clean, we don't know how all of this will react afterward. Except calling a new begin() afterward to ensure all is correctly initialized so we're back in startup point.

src/TheThingsNetwork.h
samuel-puschacher: But I think no one would put a second RN2843 on the Uno!
My Opinion: make a special constant, you have to define if you want to use a second RN2843!
No beginner would like to connect a second RN2843, the people which think about such things would also be able to make a special configuration! And for the majority of people, the uno will be power efficient.

src/TheThingsNetwork.cpp
jpmeijers: Now I understand what you are trying to do. I did the same in this repos: https://github.com/ncthompson/ThingsWeather/blob/master/firmware/things_uno/things_uno.ino
We however saw that after a few weeks the RN2483 would not wake up as expected, so we reverted back to using the ATmega's WDT.

Is it really impossible to handle the serial interrupt and re-bauding of the RN2483 from the user space, without changing the library? How about just manually doing an autobaud, including the begin() and baud changes, but in the user's program?

src/TheThingsNetwork.cpp
johanstokking: @hallard @samuel-puschacher you are changing existing code and replacing that by using `sprintf()`. That has nothing to do with this PR.

johanstokking:
@samuel-puschacher I cannot fork the repository as it is now:

1. The [commit trail](https://github.com/samuel-puschacher/arduino-boards/commits/master) is very dirty
2. The URLs should be on github.com/TheThingsProducts/... We can fork it and change it, but see [1]
3. GitHub has tags for versioning and releases for distribution, see for example the [Arduino library](https://github.com/TheThingsNetwork/arduino-device-lib/releases). We can fork it and do our own tags and releases, but then this repository has to be clean, and see also [1]
4. The product is called **The Things Uno** (no caps)

[1] The point of forking is to diverge from an existing project and/or contribute changes back to the original repository. Are you planning on maintaining this repository in your account, regardless of what is in `TheThingsProducts` (i.e. diverging projects)? If not, why not create a PR on the existing repository?

hallard:
@johanstokking 
When I first forked the library, the code was already containing some `sprintf()` in [v2.5.1](https://github.com/TheThingsNetwork/arduino-device-lib/blob/v2.5.1/src/TheThingsNetwork.cpp#L614-L628). It was done by this [commit](https://github.com/TheThingsNetwork/arduino-device-lib/commit/d5523f245e472651be2fe096562d4f642b329d60#diff-b2f2810008e6e975eb686e14fbd00f7f)
So when I optimized the code size, I won approx 700 bytes of flash, and yes I added `sprintf()` but it was not new using this call since some where already in place on original code.
I don't see what I've done wrong on this point.
Charles

johanstokking:
Indeed, it did contain `sprintf()`, but I'd rather get rid of it than introducing more occurrences, especially in unrelated feature branches like these.

jpmeijers:
Merging this into a separate branch to do the splitting.

PR #233, opened 2018-03-31
Support for IN865_867

akshaim:


src/TheThingsNetwork.cpp
```diff
@@ -748,7 +748,38 @@ void TheThingsNetwork::configureKR920_923()
   }
   sendMacSet(MAC_PWRIDX, TTN_PWRIDX_KR920_923);
 }
+void TheThingsNetwork::configureIN865_867()
+{
+  sendMacSet(MAC_ADR, "off"); // TODO: remove when ADR is implemented for this plan
+  sendMacSet(MAC_RX2, "2 866550000"); //  SF10  
+  //disable two default LoRaWAN channels
```
jpmeijers: `two` => `three`

src/TheThingsNetwork.cpp
```diff
@@ -748,7 +748,38 @@ void TheThingsNetwork::configureKR920_923()
...
+  uint8_t ch;
+  for (ch = 3; ch < 6; ch++)
+  {
+    sendChSet(MAC_CHANNEL_DCYCLE, ch, "799");
+    sprintf(buf, "%lu", freq);
+    sendChSet(MAC_CHANNEL_FREQ, ch, buf);
+    sendChSet(MAC_CHANNEL_DRRANGE, ch, "0 6");
+    sendChSet(MAC_CHANNEL_STATUS, ch, "on");
+    switch(ch)
+		{
```
jpmeijers: The indentation is a bit strange here.

akshaim:
865_867 addition was tested to work on two channels but 865985000 is showing a few issues. I shall debug and update soon.

johanstokking:
@akshaim thanks! Should we wait for your debugging and test results?

akshaim:
Sure. I am waiting for an SDR to arrive to see exactly what is going on.

akshaim:
Works now. Uplink and Downlink tested.

jpmeijers:
It is fine as it is, but I would have written this function differently as there is no need for the loop and switch - they just complicate the algorithm without saving any memory.

```
void TheThingsNetwork::configureIN865_867()
{
  sendMacSet(MAC_ADR, "off"); // TODO: remove when ADR is implemented for this plan
  sendMacSet(MAC_RX2, "2 866550000"); // SF10  
  
  // Disable the three default LoRaWAN channels
  sendChSet(MAC_CHANNEL_STATUS, 0, "off");
  sendChSet(MAC_CHANNEL_STATUS, 1, "off");
  sendChSet(MAC_CHANNEL_STATUS, 2, "off");

  // Channel 3
  sendChSet(MAC_CHANNEL_DCYCLE, 3, "799");
  sendChSet(MAC_CHANNEL_FREQ, 3, "865062500");
  sendChSet(MAC_CHANNEL_DRRANGE, 3, "0 5");
  sendChSet(MAC_CHANNEL_STATUS, 3, "on");

  // Channel 4
  sendChSet(MAC_CHANNEL_DCYCLE, 4, "799");
  sendChSet(MAC_CHANNEL_FREQ, 4, "865402500");
  sendChSet(MAC_CHANNEL_DRRANGE, 4, "0 5");
  sendChSet(MAC_CHANNEL_STATUS, 4, "on");

  // Channel 5
  sendChSet(MAC_CHANNEL_DCYCLE, 5, "799");
  sendChSet(MAC_CHANNEL_FREQ, 5, "865985000");
  sendChSet(MAC_CHANNEL_DRRANGE, 5, "0 5");
  sendChSet(MAC_CHANNEL_STATUS, 5, "on");
  
  sendMacSet(MAC_PWRIDX, TTN_PWRIDX_IN865_867);
}
```

src/TheThingsNetwork.cpp
```diff
@@ -749,6 +749,39 @@ void TheThingsNetwork::configureKR920_923()
...
+  sendChSet(MAC_CHANNEL_STATUS, 0, "off");
+  sendChSet(MAC_CHANNEL_STATUS, 1, "off");
+  sendChSet(MAC_CHANNEL_STATUS, 2, "off");
+
+  char buf[10];
+  uint32_t freq = 865062500;
+  uint8_t ch;
+  for (ch = 3; ch < 7; ch++)
+  {
+    sendChSet(MAC_CHANNEL_DCYCLE, ch, "799");
```
jpmeijers: For 1% duty cycle on 3 channels, this value should be `99`, not `799`.

```
The <dutyCycle> value that needs to be configured can be
obtained from the actual duty cycle X (in percentage)
using the following formula: <dutyCycle> = (100/X) – 1
```

akshaim: I shall try the same and update you soon.

jpmeijers: Actually now that I slept on this I doubt it again:

1% duty cycle over 3 channels is 0.333% per channel.
`(100/0.333) - 1 = 299`

So the value should be `299`.

akshaim:
Thanks. I have updated the code as per your recommendations and it works! Can you help me implement ADR on the channel?

johanstokking:
What is the reason to disable the three default LoRaWAN channels?

Regarding the loop; this is consistent with other frequency plans (and it may reduce flash memory).

johanstokking:
Any news @akshaim @jpmeijers ?

jpmeijers:
1. On the RN module with european firmware one can not change the frequency of the three default channels. For India we therefore have to disable these three channels to not use them, and rather use another three to configure the three default Indian channels.

2. Even though the original loop method is consistent with the other frequency plans, it complicates the code and in this case do not save any space. In this case there is no constant step between 8 channel, so using a loop to add the constant step doesn't make sense. Configuring the three channels one by one by hand is much clearer and does not really use much more program space than the loop. In fact, I believe it will use less instructions as we are not using the loop. For-loops are notorious to be compiled to a lot of instructions and are quite often unrolled by the compiler.

jpmeijers:
Also note that we plan to rewrite the frequency plan logic to save program space. So let's not worry too much about it now as we will rethink and re-implement it later. 
See branch:
https://github.com/TheThingsNetwork/arduino-device-lib/tree/feature/use-less-program-space

akshaim:
@johanstokking Please see the following snap from the RN2483 command reference manual. Channel 0-2 can be used only on default channels.
https://imgur.com/a/sSO9SMp

PR #234, opened 2018-04-04
Autobaud, sleep, wake changes

jpmeijers:
Let's discuss this first before emerging.

### DIFF
--- a/src/TheThingsNetwork.cpp
+++ b/src/TheThingsNetwork.cpp
@@ -287,7 +287,7 @@ uint8_t receivedPort(const char *s)
   return port;
 }
 
-TheThingsNetwork::TheThingsNetwork(Stream &modemStream, Stream &debugStream, ttn_fp_t fp, uint8_t sf, uint8_t fsb)
+TheThingsNetwork::TheThingsNetwork(SerialType &modemStream, Stream &debugStream, ttn_fp_t fp, uint8_t sf, uint8_t fsb)
 {
   this->debugStream = &debugStream;
   this->modemStream = &modemStream;
@@ -387,6 +387,12 @@ size_t TheThingsNetwork::readResponse(uint8_t prefixTable, uint8_t indexTable, u
   return readLine(buffer, size);
 }
 
+size_t TheThingsNetwork::checkModuleAvailable()
+{
+    // Send sys get ver check we have an answer
+    return readResponse(SYS_TABLE, SYS_TABLE, SYS_GET_VER, buffer, sizeof(buffer));  
+}
+
 void TheThingsNetwork::autoBaud()
 {
   // Courtesy of @jpmeijers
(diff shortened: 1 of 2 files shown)


src/TheThingsNetwork.cpp
```diff
@@ -387,6 +387,12 @@ size_t TheThingsNetwork::readResponse(uint8_t prefixTable, uint8_t indexTable, u
   return readLine(buffer, size);
 }
 
+size_t TheThingsNetwork::checkModuleAvailable()
```
jpmeijers: Having a function like this is a good idea. I would however like to change it to check the version of the RNxxxx module and return a value from an enum stating which hardware and firmware version it is.

We can then use the result to check before we set the frequency plan, and prevent errors when setting the wrong FP into the wrong module.

src/TheThingsNetwork.cpp
```diff
@@ -396,19 +402,51 @@ void TheThingsNetwork::autoBaud()
...
   modemStream->setTimeout(10000);
-  baudDetermined = true;
+}
+
+bool TheThingsNetwork::isSleeping()
+{  
+  return !baudDetermined;
+}
+
+void TheThingsNetwork::wake()
```
jpmeijers: There isn't much difference between the `wake()` and `autoBaud()` functions. I would recommend we merge these two.

hallard:
Thanks for splitting @jpmeijers , looks good to me

src/TheThingsNetwork.cpp
```diff
@@ -396,19 +402,51 @@ void TheThingsNetwork::autoBaud()
...
   }
   delay(100);
   clearReadBuffer();
   modemStream->setTimeout(10000);
-  baudDetermined = true;
+}
+
+bool TheThingsNetwork::isSleeping()
+{  
+  return !baudDetermined;
```
jpmeijers: I think we should rename this flag to something like `moduleSleeping`.

src/TheThingsNetwork.cpp
```diff
@@ -969,11 +1007,9 @@ void TheThingsNetwork::sleep(uint32_t mseconds)
   modemStream->write(buffer);
   modemStream->write(SEND_MSG);
   debugPrintLn(buffer);
-}
 
-void TheThingsNetwork::wake()
-{ 
-  autoBaud();
+  // to be determined back on wake up
+  baudDetermined = false;
```
jpmeijers: Again, `moduleSleeping` would make more sense here.

src/TheThingsNetwork.cpp
```diff
@@ -396,19 +402,51 @@ void TheThingsNetwork::autoBaud()
...
+    // check we can talk
+    length = checkModuleAvailable();
+    
+    // We succeeded talking to the module ?
+    baudDetermined = (length > 0) ;
   }
   delay(100);
   clearReadBuffer();
   modemStream->setTimeout(10000);
-  baudDetermined = true;
```
jpmeijers: I actually don't know why we added this flag to start with. Presumably to do a check before every `sendCommand` call and `autoBaud` if we didn't do so yet. If we do decide to implement this, we need to write the `autoBaud` and `checkModuleAvailable` functions not to use `sendCommand`.

But for now maybe let's just scrap the `baudDetermined` flag and rather introduce a `moduleSleeping` flag.

johanstokking:
@hallard can you address @jpmeijers review comments?

src/TheThingsNetwork.cpp
hallard: makes sense

src/TheThingsNetwork.cpp
hallard: Yeah, I do not wanted to change your original one, but I prefer also to merge both, much cleaner

src/TheThingsNetwork.cpp
hallard: ok

src/TheThingsNetwork.cpp
hallard: that's fine

src/TheThingsNetwork.cpp
jpmeijers: Good point. Maybe not a flag then but an enum. Something like this perhaps? Please suggest naming.
```
enum state {
  UNKNOWN,
  AWAKE,
  SLEEPING
}
```

hallard: sounds good, may be UNKNOWN => NOT_FOUND or NONE or MISSING

johanstokking:
Hmm, you're approving but I don't see @jpmeijers suggestions implemented.

jpmeijers:
Yes, I'll make the improvements and come back for feedback. Won't happen that fast though. Perhaps the weekend.

src/TheThingsNetwork.cpp
```diff
@@ -287,7 +287,7 @@ uint8_t receivedPort(const char *s)
   return port;
 }
 
-TheThingsNetwork::TheThingsNetwork(Stream &modemStream, Stream &debugStream, ttn_fp_t fp, uint8_t sf, uint8_t fsb)
+TheThingsNetwork::TheThingsNetwork(SerialType &modemStream, Stream &debugStream, ttn_fp_t fp, uint8_t sf, uint8_t fsb)
```
cimm: 1. Random drive by comment but don't forget to update the [documentation](https://github.com/TheThingsNetwork/arduino-device-lib/blob/master/docs/TheThingsNetwork.md#class-thethingsnetwork) since the `Stream` is changing to a `SerialType`.
2. Why change the modemStream to a `SerialType` but not the debugStream?
3. Wouldn't it make sense to change the variable name from `modemStream` to `modemSerial` or something in that style to avoid confusion?

johanstokking:
@cimm you may want to assist @jpmeijers if wanted!

cimm:
@johanstokking Happy to help but... I am a software developer, not (yet) a hardware guy. I can see `SerialType &modemStream` doesn't make sense but I have no idea why we go for a SerialType over a Stream. [Simple renaming](https://github.com/cimm/arduino-device-lib/commit/5f51a36b766e057e51c6e660aa57aabccfce3d26) I can do. :smiley: 

Happy to do the work if someone can give me some simple beginner tasks.

PR #240, opened 2018-07-17
Fix TheThingsMessage documentation links

cimm:
I noticed 2 broken links in the TheThingsNetwork documentation. The links to the `Send.ino` and `ABP.ino` examples are broken. It looks like the filenames changed. I fixed those.

Also changed the absolute URLs to relative ones since the absolute links would point to the TheThingsNetwork repository even if you are on a forked or offline repository.

Tiny changes.

### DIFF
--- a/docs/TheThingsNetwork.md
+++ b/docs/TheThingsNetwork.md
@@ -64,7 +64,7 @@ RX Delay 1: 1000
 RX Delay 2: 2000
 ```
 
-See the [DeviceInfo](https://github.com/TheThingsNetwork/arduino-device-lib/blob/master/examples/DeviceInfo/DeviceInfo.ino) example.
+See the [DeviceInfo](../examples/DeviceInfo/DeviceInfo.ino) example.
 
 ## Method: `onMessage`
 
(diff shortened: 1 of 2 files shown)


docs/TheThingsNetwork.md
```diff
@@ -64,7 +64,7 @@ RX Delay 1: 1000
 RX Delay 2: 2000
 ```
 
-See the [DeviceInfo](https://github.com/TheThingsNetwork/arduino-device-lib/blob/master/examples/DeviceInfo/DeviceInfo.ino) example.
+See the [DeviceInfo](../examples/DeviceInfo/DeviceInfo.ino) example.
```
johanstokking: Please leave the absolute paths; we're also importing this Markdown in external documentation that are unable to handle the relative path.

cimm:
@johanstokking Thanks for the [feedback](https://github.com/TheThingsNetwork/arduino-device-lib/pull/240#discussion_r203759632), reverted the relative URLs to the absolute Github ones.

docs/TheThingsNetwork.md
cimm: Thanks for the feedback. I reverted the relative URLs to the absolute Github ones.

Issue #242, opened 2018-08-10
Implement error handling/timeout on readLine from LoRa module


savnik:
I sometimes experience that the LoRa module(RN2483) does not send anything back after a sendBytes. The current implementation will wait forever until a return signal is send. 

`size_t TheThingsNetwork::readLine(char *buffer, size_t size)
{
  size_t read = 0;
  while (read == 0)
  {
    read = modemStream->readBytesUntil('\n', buffer, size);
  }
  buffer[read - 1] = '\0'; // set \r to \0
  return read;
}` 

I propose that a timeout is implemented to handle if there is no response from RN2483 within reasonable time.

savnik:
Yeah I see your point about returning too early. RN module as you say should work 100% of the time, however sometimes we see it locks up, not knowing why or what triggered it. I would agree that fixing the root cause would be the best solution, but unfortunately we cant access the raw firmware of the RN module to figure out what goes wrong. So then we though about how to handle this error.

If the function is returning a zero, as you suggest, it will give the application the possibility to handle the error which is an enhancement to now. 
Have you been working on such a solution already?

savnik:
I see two approaches to handle the timeout. 
1. Using milis() to measure the time spend in the while loop. 
```
size_t TheThingsNetwork::readLine(char *buffer, size_t size, uint8_t timeout = 10000) // Default timeout vallue TBD
{
  size_t read = 0;
  uint8_t start_time = milis();
  while (read == 0 && milis()-start_time<timeout)
  {
    read = modemStream->readBytesUntil('\n', buffer, size);
  }
  if(milis()-start_time<timeout)){ // If timeout is activated return 0 to inform the application
    return 0;
  }
  else{
    buffer[read - 1] = '\0'; // set \r to \0
    return read;
  }
}
```

2. Using modemStream timeout. This assumes that somewhere we set modemStream->setTimeout(x). Then if x = 1, then we will achive approximately 10s timeout as in option 1. 
```
size_t TheThingsNetwork::readLine(char *buffer, size_t size, uint8_t attempts = 10) // Default timeout attempts TBD
{
  size_t read = 0;
  while (read == 0 && attempts>0)
  {
    read = modemStream->readBytesUntil('\n', buffer, size);
    attempts--;
  }
  if(attempts<=0){ // If timeout attempts is activated return 0 to inform the application
    return 0;
  }
  else{
    buffer[read - 1] = '\0'; // set \r to \0
    return read;
  }
[...]

PR #244, opened 2018-08-17
Adding attempts to readLine and state marker if radio module is unres…

savnik:
Proposal to fix issue https://github.com/TheThingsNetwork/arduino-device-lib/issues/242

### DIFF
--- a/docs/TheThingsNetwork.md
+++ b/docs/TheThingsNetwork.md
@@ -31,7 +31,7 @@ void reset(bool adr);
 Performs a hardware reset of the RN module. Input parameter is the pin which the reset pin from the module is connected to. This does clear saved state, e.g. provisioned keys.
 
 ```c
-void hardResetuint8_t resetPin);
+void hardReset(uint8_t resetPin);
 ```
 
 - `bool adr`: Enable/disable Adaptive Data Rate.
--- a/docs/TheThingsNetwork.md
+++ b/docs/TheThingsNetwork.md
@@ -31,6 +31,7 @@ void reset(bool adr);
 ## Method: `hardReset`
 
 Performs a hardware reset of the RN module. Input parameter is the pin which the reset pin from the module is connected to. This does clear saved state, e.g. provisioned keys.
+The `resetPin` should be configured as output and set to high.
 
 ```c
 void hardReset(uint8_t resetPin);
--- a/docs/TheThingsNetwork.md
+++ b/docs/TheThingsNetwork.md
@@ -31,13 +31,12 @@ void reset(bool adr);
 ## Method: `hardReset`
 
 Performs a hardware reset of the RN module. Input parameter is the pin which the reset pin from the module is connected to. This does clear saved state, e.g. provisioned keys.
-The `resetPin` should be configured as output and set to high.
 
 ```c
 void hardReset(uint8_t resetPin);
 ```
 
-- `uint8_t resetPin`: The output pin that is connected to the module's reset pin.
+- `uint8_t resetPin`: The output pin that is connected to the module's reset pin. The output pin should be configured as output and set to high by the user.
 
 ## Method: `getHardwareEui`
 
--- a/docs/TheThingsNetwork.md
+++ b/docs/TheThingsNetwork.md
@@ -26,6 +26,14 @@ Performs a software reset of the RN module. This does not clear saved state, e.g
 void reset(bool adr);
 ```
 
+## Method: `hardReset`
+
+Performs a hardware reset of the RN module. Input parameter is the pin which the reset pin from the module is connected to. This does clear saved state, e.g. provisioned keys.
+
+```c
+void hardResetuint8_t resetPin);
+```
+
 - `bool adr`: Enable/disable Adaptive Data Rate.
 
 ## Method: `getHardwareEui`
--- a/docs/TheThingsNetwork.md
+++ b/docs/TheThingsNetwork.md
@@ -36,7 +36,7 @@ Performs a hardware reset of the RN module. Input parameter is the pin which the
 void hardReset(uint8_t resetPin);
 ```
 
-- `uint8_t resetPin`: The output pin that is connection to the module's reset pin.
+- `uint8_t resetPin`: The output pin that is connected to the module's reset pin.
 
 ## Method: `getHardwareEui`
 
--- a/docs/TheThingsNetwork.md
+++ b/docs/TheThingsNetwork.md
@@ -26,6 +26,8 @@ Performs a software reset of the RN module. This does not clear saved state, e.g
 void reset(bool adr);
 ```
 
+- `bool adr`: Enable/disable Adaptive Data Rate.
+
 ## Method: `hardReset`
 
 Performs a hardware reset of the RN module. Input parameter is the pin which the reset pin from the module is connected to. This does clear saved state, e.g. provisioned keys.
@@ -34,7 +36,7 @@ Performs a hardware reset of the RN module. Input parameter is the pin which the
 void hardReset(uint8_t resetPin);
 ```
 
-- `bool adr`: Enable/disable Adaptive Data Rate.
+- `uint8_t resetPin`: The output pin that is connection to the module's reset pin.
 
 ## Method: `getHardwareEui`
 
--- a/src/TheThingsNetwork.cpp
+++ b/src/TheThingsNetwork.cpp
@@ -357,15 +357,21 @@ void TheThingsNetwork::clearReadBuffer()
   }
 }
 
-size_t TheThingsNetwork::readLine(char *buffer, size_t size)
+size_t TheThingsNetwork::readLine(char *buffer, size_t size, uint8_t attempts = 3) // Default timeout value 10s and is set in class initiator
 {
   size_t read = 0;
-  while (read == 0)
+  while (read == 0 && attempts>0)
   {
     read = modemStream->readBytesUntil('\n', buffer, size);
   }
-  buffer[read - 1] = '\0'; // set \r to \0
-  return read;
+  if(attempts<=0){ // If attempts is activated return 0 and set RN state marker
+    this->radioModuleInvalidState = true; // Inform the application about the radio module is not responsive.
+    return 0;
+  }
+  else{
+    buffer[read - 1] = '\0'; // set \r to \0
+    return read;
+  }
 }
 
 size_t TheThingsNetwork::readResponse(uint8_t prefixTable, uint8_t index, char *buffer, size_t size)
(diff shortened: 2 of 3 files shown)


src/TheThingsNetwork.cpp
```diff
@@ -357,15 +357,21 @@ void TheThingsNetwork::clearReadBuffer()
...
   {
     read = modemStream->readBytesUntil('\n', buffer, size);
   }
-  buffer[read - 1] = '\0'; // set \r to \0
-  return read;
+  if(attempts<=0){ // If attempts is activated return 0 and set RN state marker
+    this->radioModuleInvalidState = true; // Inform the application about the radio module is not responsive.
+    return 0;
+  }
+  else{
```
johanstokking: Please make spacing consistent around braces and brackets. Also no need for `else` after a `return`.

src/TheThingsNetwork.cpp
```diff
@@ -357,15 +357,21 @@ void TheThingsNetwork::clearReadBuffer()
   }
 }
 
-size_t TheThingsNetwork::readLine(char *buffer, size_t size)
+size_t TheThingsNetwork::readLine(char *buffer, size_t size, uint8_t attempts = 3) // Default timeout value 10s and is set in class initiator
```
johanstokking: Where is it set?

savnik: timeout is set in:
https://github.com/TheThingsNetwork/arduino-device-lib/blob/4629038364b748452bad617d008442f15e948238/src/TheThingsNetwork.cpp#L294

src/TheThingsNetwork.cpp
savnik: Fixed

savnik:
@johanstokking did you have time to review the changes?

src/TheThingsNetwork.cpp
johanstokking: Set defaults in the header file

src/TheThingsNetwork.cpp
```diff
@@ -357,13 +357,18 @@ void TheThingsNetwork::clearReadBuffer()
...
 {
   size_t read = 0;
-  while (read == 0)
+  while (read == 0 && attempts>0)
```
johanstokking: Spacing; `attemps > 0`

johanstokking: Shouldn't you be decrementing `attempts` somewhere? Like `while (!read && attempts--)` ?

src/TheThingsNetwork.cpp
```diff
@@ -357,13 +357,18 @@ void TheThingsNetwork::clearReadBuffer()
...
   {
     read = modemStream->readBytesUntil('\n', buffer, size);
   }
+  if(attempts<=0)
+  { // If attempts is activated return 0 and set RN state marker
+    this->radioModuleInvalidState = true; // Inform the application about the radio module is not responsive.
```
johanstokking: Although this is a way to signal, still it requires action from the application. I don't really know if application developers know how to resolve this.

The best way to resolve this is to reset the RN module through the reset pin. Would it be a good idea to configure (optionally) the reset pin, and engage a soft reset in the library at this point?

src/TheThingsNetwork.h
```diff
@@ -54,10 +54,11 @@ class TheThingsNetwork
   bool adr;
   char buffer[512];
   bool baudDetermined = false;
+  bool radioModuleInvalidState = false;
   void (*messageCallback)(const uint8_t *payload, size_t size, port_t port);
 
   void clearReadBuffer();
-  size_t readLine(char *buffer, size_t size);
+  size_t readLine(char *buffer, size_t size, uint8_t attempts);
```
johanstokking: So add the default always in the header file

src/TheThingsNetwork.cpp
savnik: I see your valid point. Now a good question is should we let the library handle the reset or give the decision to the application. If the library is resetting the RN module we will need to let the application know so the application can handle this, eg. by re-joining the network.
But in either way we must raise a flag to the application and abort the current task. 
If we reset the RN module in the library we can do a soft-reset by sending a sys_reset cmd or a hard-reset using the reset pin. In either way the library needs to rejoin the network and then re-execute the function. Should the module then retry for infinity or a limited tries?

One option could be to soft/hard reset the module every time this error occur, print a debug message with details and raise the flag. In this way we have raised awareness of a problem and tried to resolve it while giving power over the exception handling to the application.

readLine is used in:
- readResponse
- personalize
- join
- sendBytes
- waitForOk

savnik: However maybe the simple solution is the best. On error exit with debug message and error flag. Then update documentation and make a good example on how to handle the error on application level.

johanstokking: My wording was wrong; I meant hard reset, by using the reset pin. Sending the soft reset command is not going to work as the serial is broken already.

And you're right, doing a hard reset in the midst of a join or TX operation, and restoring the state (i.e. joining again) is not going to work either.

So I'm good with the signaling to the application. Please rename to `needsHardReset`, introduce a `resetHard()` that takes a pin number as parameter, and reset the `needsHardReset` after hard resetting. Also documentation would be helpful. Thanks!

savnik: Okay. I will work on this and come back soon.

src/TheThingsNetwork.cpp
```diff
@@ -436,6 +443,17 @@ void TheThingsNetwork::reset(bool adr)
...
   this->adr = adr;
+  this->needsHardReset = false;
+}
+
+void TheThingsNetwork::resetHard(uint8_t resetPin){
+  digitalWrite(resetPin, LOW);
+  unsigned long time_now = milis();
+  int period = 1000;
+  while(millis() < time_now + period){
+    // wait
```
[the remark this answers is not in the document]
savnik: It was just to make it non-blocking in case the application is using interrupt. The reset function is not time critical. 
I have looked in the datasheet and there is no advice on minimum time for the signal for a reset to be triggered. I know that 1s works by test, but I have not tested <1s.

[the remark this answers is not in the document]
jpmeijers: Actually delay() is the better one to use. On some platforms where it is necessary delay() will call yield(), while a while loop will not.

docs/TheThingsNetwork.md
```diff
@@ -26,6 +26,14 @@ Performs a software reset of the RN module. This does not clear saved state, e.g
 void reset(bool adr);
 ```
 
+## Method: `hardReset`
+
+Performs a hardware reset of the RN module. Input parameter is the pin which the reset pin from the module is connected to. This does clear saved state, e.g. provisioned keys.
+
+```c
+void hardResetuint8_t resetPin);
```
johanstokking: Missing parentheses

docs/TheThingsNetwork.md
```diff
@@ -26,6 +26,14 @@ Performs a software reset of the RN module. This does not clear saved state, e.g
...
+```
+
 - `bool adr`: Enable/disable Adaptive Data Rate.
```
johanstokking: This parameter belongs to `reset()`, above

src/TheThingsNetwork.cpp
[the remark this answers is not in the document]
johanstokking: Can you implement this @savnik ?

savnik:
So the wait function is implemented. I have tried to make it more clear now. 
Maybe we do not see the same regarding spacing - but i can not see where the spacing is wrong now?

jpmeijers:
Only saw the review request now. I'll try and give this PR a proper look and review during the coming week.

src/TheThingsNetwork.cpp
```diff
@@ -436,6 +442,18 @@ void TheThingsNetwork::reset(bool adr)
...
+
+void TheThingsNetwork::resetHard(uint8_t resetPin){
+  digitalWrite(resetPin, LOW);
+  unsigned long time_now = millis();
+  int period = 1000;
+  // Non I/O blocking wait function
+  while(millis() < time_now + period)
+  {
+  }
+  digitalWrite(resetPin, HIGH);
```
jpmeijers:
```
  digitalWrite(resetPin, LOW);
  delay(1000);
  digitalWrite(resetPin, HIGH);
```

We need to use delay, and not write our own delay loop. This is so that this code won't crash the wifi on the ESP8266. Alternatively one can add a `yield()` in the body of the while, but not all platforms has the yield() function. So better use `delay()`.

docs/TheThingsNetwork.md
```diff
@@ -28,6 +28,16 @@ void reset(bool adr);
...
+## Method: `hardReset`
+
+Performs a hardware reset of the RN module. Input parameter is the pin which the reset pin from the module is connected to. This does clear saved state, e.g. provisioned keys.
+
+```c
+void hardReset(uint8_t resetPin);
+```
+
+- `uint8_t resetPin`: The output pin that is connected to the module's reset pin.
+
```
jpmeijers: We are assuming here that the user already did a pinMode(resetPin, OUTPUT); previously in their code. Let's add to the documentation that the resetPin that is passed to the `hardReset()` function should be a configured output pin.

src/TheThingsNetwork.cpp
savnik: I have not worked with delay functions on ESP8266, but i would assume that using delay function it would block the wifi operation, just as the delay function on a "normal" arduino blocks the interrupt. 
Can you explain why the wifi will crash using this delay loop?

docs/TheThingsNetwork.md
savnik: Good point here. Will add this.

src/TheThingsNetwork.cpp
jpmeijers: "the delay function on a "normal" arduino blocks the interrupt" - I'm not sure that this is true. A delay inside an ISR will prevent further interrupts to trigger, but this is not true in the normal execution path. So a delay in normal code will be interrupted just like any other code.

But let me reiterate this:
Do not write custom implementations where perfectly working functions are given by the platform. Different platforms handle things differently lower down, and the abstraction that Arduino gives us should be used to prevent thing from going wrong. The ESP8266 example is just one of many possible situations where a custom blocking delay loop will cause issues with lower down system services.

On https://learn.sparkfun.com/tutorials/esp8266-thing-hookup-guide/using-the-arduino-addon, read the paragraph under Yielding.

ElectronicallyE:
Thanks guys on making progress on this issue. I was about to write up an issue about this, but seems like you've got a handle on it already.

src/TheThingsNetwork.cpp
savnik: Thanks, it is pretty clear to me now. I will update my PR.

savnik:
@jpmeijers have you had time to take a look at the two changes you sugested?

johanstokking:
Released in https://github.com/TheThingsNetwork/arduino-device-lib/releases/tag/v2.5.14

Thanks @savnik and @jpmeijers

PR #250, opened 2019-01-20
Avoid compiler warning

AmedeeBulle:
This is more cosmetic than anything else, but I try to avoid unnecessary compiler warnings in builds.  
The `unused` attributes acknowledges the variable is unused and suppress the warning.

Alternatively we could cast `readResponse()` to `(void)` instead of using the `length` variable...


johanstokking:
Thanks!

How about not declaring `size_t length` in the first place?

AmedeeBulle:
That's the alternative I was suggesting as well...

Updated the MR.

PR #258, opened 2019-07-14
pb.h should be included from the local directory

jpmeijers:
`pb.h` should be an include from the local directory, therefore `#include "ph.h"` and not `#include <pb.h>`.

Compile will fail if this library is not being included from the Arduino global path, but rather from the local directory. For example if the directory structure looks like this:

```
example/
- example.ino
- src/
-- arduino-device-lib/
```

and inside example.ino we include the library as follows:
```
#include "src/arduino-device-lib/src/TheThingsNetwork.h"
```

After this change the compile is successful.

Note:
If the library is present in both the Arduino libraries directory as well as the local directory, compilation might fail due to duplicate definitions. Keep this in mind during testing.

### DIFF
--- a/src/TheThingsMessage.h
+++ b/src/TheThingsMessage.h
@@ -5,9 +5,9 @@
 #define _THETHINGSMESSAGE_H_
 
 #include "TheThingsNetwork.h"
-#include <pb.h>
-#include <pb_encode.h>
-#include <pb_decode.h>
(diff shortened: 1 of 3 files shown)


jpmeijers:
Strange. Seems like I had to re-authorize the CLA Assistant with Github OAuth2 for it to pick up my previous signed CLA.

PR #260, opened 2020-04-19
Class C support

### LINKED ISSUE #259: Class C support?
Hi there, is Class C downlink supported by this library? If no, do you have plan to support it? 
Thanks for the library by the way!

jpmeijers:
Fixes #259 

Tested against [The Things Stack](https://github.com/TheThingsNetwork/lorawan-stack) V3.7.0


examples/ReceiveClassC/ReceiveClassC.ino
```diff
@@ -0,0 +1,62 @@
...
+  // Class C RX only takes affect after a TX
+  uint8_t payload[] = {0x00};
+  ttn.sendBytes(payload, 1);
+}
+
+void loop()
+{
+  debugSerial.println("-- LOOP");
+
+  // Send single byte to poll for incoming messages
```
johanstokking, line 44: This comment is not accurate for class C; we're not sending bytes but reading from the incoming stream

src/TheThingsNetwork.cpp
```diff
@@ -603,8 +621,44 @@ ttn_response_t TheThingsNetwork::sendBytes(const uint8_t *payload, size_t length
 
 ttn_response_t TheThingsNetwork::poll(port_t port, bool confirm)
 {
-  uint8_t payload[] = {0x00};
-  return sendBytes(payload, 1, port, confirm);
+  if(lw_class == CLASS_C) {
```
johanstokking, line 624: nit; I would use `switch`/`case { ... }` here

johanstokking:
@jpmeijers thanks for updating, please request review when ready!

PR #270, opened 2020-10-22
Add files via upload - Firmware AU915 V1.0.5

hjmhardsoft:
Firmware version 1.0.5 adapted to our frequency plan (AU-915), which works in class C. As it is not the final version, keep in mind that it can sometimes throw mac_error when transmitting if the ADR is activated.


hjmhardsoft:
I have tested this Firmware with Sodaq Explorer and TTN in Argentina using Meshed-Router.

hjmhardsoft:
Firmware version 1.0.5 adapted to our frequency plan (AU-915), which works in class C. As it is not the final version, keep in mind that it can sometimes throw mac_error when transmitting if the ADR is activated.

hjmhardsoft:
Firmware version 1.0.5 adapted to our frequency plan (AU-915), which works in class C. As it is not the final version, keep in mind that it can sometimes throw mac_error when transmitting if the ADR is activated.

hjmhardsoft:
Someone can review ? This firmware added is necesary for Argentina/Australia for Example.
I don´t know more updated firmware for RN2903 under AU915.

PR #277, opened 2021-12-10
Feature upgrades for ABP, add errorReadout, poll after Retry Class A and more

flhofer:
Hello!
[EDITED]

I made some changes to the library for a project I worked on, so I created a pull request for the most useful ones.

## Changes in general function.

- sendBytes returns now `UNSUCCESSFUL_RECEIVE` if the sent frame message is confirmed, but no response has been received
- poll now includes a flag `modem_only` to only listen to the modem's output. Useful to poll again if a confirmed message does not get through and the modem retires in the background.
- poll added a check for receive length. If length is 0, we get a `UNSUCCESSFUL_RECEIVE` and not `SUCCESSFUL_RECEIVE`.

## New methods

- getters for Bandwidth, CodeRate, Frequency, DR, WDT, Power, PowerIndex, RSSI, SNR, FCU and FCD frame counters
- setters for LoRaWan power index (max TX power), data rate, FCU and FCD frame counters; Useful for ABP manual operation and when restoring from EEPROM/Flash
- getStatus to read the modem status, e.g, sending, RX1 window open, ecc.. (requires Firmware >= 1.05)
- getLastError to read and return the last error code in the receive buffer. Adds also an enumeration and text constants
- added methods setChannel, setRx2Channel, and SetChannelStatus to configure, enable and disable channels. Useful for the configuration of channels 4-8 when running ABP and recovering status from EEPROM.
- setChannelDCycle to modify a channel's duty cycle. This might be useful if you don't use all default channels of a band, thus can increase the remaining channel's DC
- setRX1Delay to modify the waiting time between TX and the first window opening. May be needed on custom Network servers when operating in ABP.

## Other changes

- added a constant `TTN_DEFAULT_TIMEOUT` as a default timeout value
- minor additions such as init values for private variables, range and size adjustments for arrays..
- Added private methods to standardize response parsing parseBytes, and setters for modems with more parameters to reduce the number of strings, - i.e., total number of bytes needed -, for the implementation of the library

## Final note 

I tested the code with a TTN Uno board on a private LoRaWan server. There may be some typos as I had to manually apply some changes from my version as it holds way more features. And yes: I will push docs and examples to the branch while this request gets rolling. I know it takes some time for review, and it's better to get started soon. 😃


jpmeijers:
The overall quality of this PR looks very good to me. Two small issues:
* Typo in define `TTN_UNSUCESSFUL_RECEIVE`, but I see that was my mistake in a previous PR. Let's fix it in a different PR than this one.
* I do not like abbreviations like `mdmonly`. Let's rather be more descriptive like `modem_only`. Also with `TTN_DEFAULT_TOUT`=>`TTN_DEFAULT_TIMEOUT`, `pos`=>`position`, `drmin`=>`dr_min`, `idx`=>`index`, `CMPERR_TABLE`=>`COMPARE_ERROR_TABLE`, etc

flhofer:
@jpmeijers Thanks for the quick reply! Didn't expect that. I have more proposals.
I will apply your proposed changes and upload also some documentation soon. Some I noticed too

flhofer:
@jpmeijers 
Your comment on `TTN_UNSUCESSFUL_RECEIVE` made me recheck the code. I actually found that too. Indeed, the submission I made yesterday was onle the changes of last year. I added thus now the rest and made a first documentation. I might still do some refactoring.. 😃 

More updates soon

flhofer:
@jpmeijers 
Refactoring done.
Also fixed a duty cycle preset issue for EU868. The default 8 channels span over 2 bands, G1 and G2, both set to 1%. Thus each of them may use 1/3 for channel 1-3 or 1/5th for channel 4-8 of the band's duty cycle of 1%, not 1/8 of 1% as it was until now.

[Explanations here](https://www.thethingsnetwork.org/docs/lorawan/duty-cycle/)

jpmeijers (approved):
I ran through the examples on my Things Uno and they work as expected.

LGTM

@johanstokking I'm happy, so you're welcome to merge if you're happy too.

src/TheThingsNetwork.cpp
```diff
@@ -700,23 +919,25 @@ void TheThingsNetwork::showStatus()
 
 void TheThingsNetwork::configureEU868()
 {
-  sendMacSet(MAC_RX2, "3 869525000");
+  sendMacSet(MAC_RX2, 3, 869525000);
```
johanstokking, line 922: I wonder if this was ever correct for OTAA. TTN V2 and V3 use SF12 for join-accept in RX2, and configure SF9 (for TTN frequency plans) through the join-accept.

In case of OTAA, the RX2 DR and which channels are enabled should not be left to default settings. Only in ABP this should be configured. I believe that `configureEU868()` is called also for OTAA, so that might be an issue. Right?

src/TheThingsNetwork.cpp
```diff
@@ -731,15 +952,15 @@ void TheThingsNetwork::configureUS915(uint8_t fsb)
   {
     if (ch == ch500 || (ch <= chHigh && ch >= chLow))
```
johanstokking, line 953: Same here I think; in OTAA, the end device should try sending join-requests in all FSBs, and the actual channels are conveyed as part of the join-accept and subsequent MAC commands.

johanstokking (commented):
Great stuff. I didn't test but I trust @jpmeijers tests.

Two questions though, while we're at it.

src/TheThingsNetwork.cpp
flhofer: Agree, but we need to keep defaults for up and down, even in OTAA

src/TheThingsNetwork.cpp
flhofer: Hi @johanstokking 
Yes, I agree with you. I didn't actually change that part, just added a private overload with one more parameter to save some program bytes here and there.

Actually, this line always confused me. The LoRa regional specs 1.0.2 define channel 9 (869.525) as default SF12. However, the same specification defines the downlink for Class B beacons @ 869.525 SF9 (DR3). I always thought it might have to do with the TTN server's default configuration (I used LorIoT so far)

Well, default join behavior defines confirmed messages on one of the three default frequencies (868.1, 868.3, and 868.5 MHz) and response within 5 or 6 seconds on RX1 or RX2 respectively. So if RX2 is wrong, it might work anyway. However, now that I added a method to change RX2, we should put that to default DR0 (Don't know the behavior of the modem if you give it nothing!). So a good point. 

So, should I make this change too? We can still configure everything like it's done now, but things like RX2 must stay default for OTAA to work as expected. Exceptional cases can use setRx2Channel.

johanstokking: Thanks for your reply. Indeed I think we should use RX2 DR of 0 in case of OTAA. The server will configure the desired RX2 DR through the join-accept. For ABP, it needs to stay DR 3.

As for Class B beaconing vs RX2; those can be different indeed. I think that DR 3 was chosen because of a larger maximum payload.

src/TheThingsNetwork.cpp
johanstokking: How do you mean?

jpmeijers:
Looking at the commands this library sends to the RN2483 when configured for EU868, I agree it does not look exactly as we want. The library uses the same `configureChannels()` for both OTAA and ABP. I think this needs to change. In the case of OTAA only the LoRaWAN default channels for the region should be configured, as the rest will be pushed by the network in the Join Accept. Only in the case of ABP we should manually configure all the channels and their duty cycles.

This is however a bigger change, and not a simple fix we should do in this PR.

@johanstokking let's merge this PR first, then we can work on fixing the RX2 window, and other issues with the configured channels.

johanstokking:
> @johanstokking let's merge this PR first, then we can work on fixing the RX2 window, and other issues with the configured channels.

I agree. Thanks a lot @flhofer for implementing and @jpmeijers for reviewing.

src/TheThingsNetwork.cpp
flhofer: @johanstokking for example in the IN865_867 band, the default channels 1-3 are disabled and others overwritten. I don't know how OTAA join works there, but it might need a different default configuration to work. Just guessing.

Issue #280, opened 2022-01-14
Handling of needsHardReset flag


danalvarez:
In my application, I am interfacing with an RN2903 module at 9600 baud. When calling `setup()`, I am checking comm with the RN module via a call to `reset()`. However, looking into the code for this function, it returns by setting `this->needsHardReset = false` at the very end:
https://github.com/TheThingsNetwork/arduino-device-lib/blob/f0a4ddd721557625273fa83e92f423970eee80a0/src/TheThingsNetwork.cpp#L650
This means that, even if comm with the module is unsuccessful, by the time the function returns, `needsHardReset` is `false` again and my application cannot know that the module is unresponsive.

My code looks like this:
```
  do {
    DEBUG_PORT.println(F("Checking comm..."));
    myLora.resetHard(RFRESET);                            // hard-reset lora module
    delay(100);                                           // wait for module startup
    // when myLora.reset(false) returns, needsHardReset
    // will be false again, even if the module was
    // unresponsive, so the do-while will exit thinking
    // comm with the module is OK
    myLora.reset(false);                                  // soft reset and adr set to false
    retries--;
  } while(myLora.needsHardReset && retries > 0);
```

I would propose removing the  `this->needsHardReset = false` in the `reset()` function, and instead setting the flag in `readLine()`, as follows:
```
size_t TheThingsNetwork::readLine(char *buffer, size_t size, uint8_t attempts)
{
  size_t read = 0;
  while (!read && attempts--)
  {
    read = modemStream->readBytesUntil('\n', buffer, size);
  }
  if (!read)
  { // If attempts is activated return 0 and set RN state marker
    this->needsHardReset = true; // Inform the application about the radio module is not responsive.
    debugPrintLn("No response from RN module.");
    return 0;
  }
  this->needsHardReset = false; // ADDED LINE: Inform the application that the radio module is responsive.
  buffer[read - 1] = '\0'; // set \r to \0
  return read;
}
```

In this manner, the same function that sets `needsHardReset` to `true` when the module is unresponsive, also sets the flag to `false` when it is responsive.

What do you think?

danalvarez:
Closing this because #282 was merged

Issue #216, opened 2017-06-22
Timed wait for response after resetting the module


johanstokking:
To know whether there's actually a working module wired and the serial has begun communicating.

See https://github.com/TheThingsNetwork/arduino-device-lib/pull/210#issuecomment-310416492

jpmeijers:
A good way would be to call `sys get ver` at statup, and checking if the response is indeed one of the modules the library supports (RN2483, RN2483A, RN2903, RN2903AS).

I'm not sure where is the best place to put this. It feels like something to do directly after the library class was instantiated. In the constructor is not possible because the Stream object was not `begin()`-ed yet at that stage. The library also doesn't have a single initialisation function. Some of our examples call `status()` at the start.

That leaves us to have a timeout on the first call to readLine(), and then setting a static flag disabling further timeouts. This however doesn't seem like an elegant solution.

Any other suggestions?

johanstokking:
How about in `reset()` ?

jpmeijers:
`reset()` is only called at the beginning of `personalize()` and `join()`. If communication with the radio doesn't work, and a user calls `showStatus()`, the library will lock up.

Communication should be checked at the beginning of all public functions that can be called first. Someone might for example call `getHardwareEui()` at the beginning.

Maybe we should add a function `boolean checkHWVersion()`, which will return false if the response from the radio is invalid. Then we should call this at the start of all public functions, returning directly with an appropriate error in case of failure. This does however cause unnecessary overhead.

johanstokking:
Right, then maybe an `init()` function is better. That would do both [this](https://github.com/TheThingsNetwork/arduino-device-lib/blob/master/src/TheThingsNetwork.cpp#L404-L407) and this connectivity check.

PR #282, opened 2022-01-17
Improve handling of needsHardReset flag and add method to check valid module connected to serial port

danalvarez:
This PR is related to #280 . I believe the handling of the `needsHardReset` flag could be improved, by not setting it to `false` with each call to `reset()`.

Additionally, I added a function called `checkValidModuleConnected()` to address https://github.com/TheThingsNetwork/arduino-device-lib/issues/216#issuecomment-310430834 . In this manner, the user application can call `checkValidModuleConnected()` in the `setup()` function of the Arduino code, and make sure a valid module is connected before any join attempt. This is especially important, because we need a function that returns as fast as possible when a module is not connected or incorrectly wired. For example, a call to `reset()` takes several minutes to return if a module is not connected, because of the many instructions carried out within it and the innate timeout given in `readLine()`. The `checkValidModuleConnected()` function returns in ~30s if `autobaud_first` is set to `false`, and ~60s if set to `true`.

I also added an example [CheckModule.ino](https://github.com/TheThingsNetwork/arduino-device-lib/blob/master/examples/CheckModule/CheckModule.ino) to showcase the usage of the `checkValidModuleConnected()` function.

Additionally, I added a `bool reset_first` to `personalize()` and `provision()`, so the user application can tell the library if they want a soft reset to be performed when calling these functions. For example, if a user performs:
1. Hard reset RN module
2. call `checkValidModuleConnected()`

Before calling `personalize()`, it would make sense to go directly into the ABP join, without a soft reset first, as the hard reset was already performed. Anyways, I set these params `true` by default so that prior behaviour is not broken.

Finally, I also updated the documentation.

### Help wanted:
**Broken:** The only caveat with this PR is that the module validation in `checkValidModuleConnected()` is not working as expected. This section in particular:

```c
  // buffer contains "RN2xx3[xx] x.x.x ...", getting only model (RN2xx3[xx])
  char *model = strtok(buffer, " ");
  debugPrintIndex(SHOW_MODEL, model);
  // check if module is valid (must be RN2483, RN2483A, RN2903 or RN2903AS)
  if(pgmstrcmp(model, CMP_RN2483) == 0 || pgmstrcmp(model, CMP_RN2483A) == 0 || pgmstrcmp(model, CMP_RN2903) == 0 || pgmstrcmp(model, CMP_RN2903AS) == 0)
  {
    debugPrintMessage(SUCCESS_MESSAGE, SCS_VALID_MODULE);
    return true;                                                // module responded and is valid (recognized/supported)
  }
```

For example, if `model = "RN2903AX"` (impossible, I know, but for the sake of the example), the function will think that this is a valid module, because `pgmstrcmp(model, CMP_RN2903) == 0` returns `true`. I am not so familiar with the c string functions. Could anyone help me by suggesting how to make the verification work correctly?

I went ahead and submitted the PR with this caveat, so that it may be reviewed and a fix to this small issue could be suggested.

### DIFF
--- a/docs/TheThingsNetwork.md
+++ b/docs/TheThingsNetwork.md
@@ -54,6 +54,14 @@ Gets the provisioned AppEUI. The AppEUI is set using `provision()` or `join()`.
 size_t getAppEui(char *buffer, size_t size);
 ```
 
+## Method: `getVersion`
+
+Gets the response from a `sys get ver` command (i.e. the hardware model, the the software version, etc.).
+
+```c
+size_t getVersion(char *buffer, size_t size);
+```
+
 ## Method: `showStatus`
 
 Writes information about the device and LoRa module to `debugStream`.
@@ -114,13 +122,14 @@ Call the method without the first two arguments if the device's LoRa module come
 Activate the device via ABP.
 
 ```c
-bool personalize(const char *devAddr, const char *nwkSKey, const char *appSKey);
+bool personalize(const char *devAddr, const char *nwkSKey, const char *appSKey, bool reset_first);
 bool personalize();
 ```
 
 - `const char *devAddr`: Device Address assigned to the device.
 - `const char *nwkSKey`: Network Session Key assigned to the device for identification.
 - `const char *appSKey`: Application Session Key assigned to the device for encryption.
+- `bool reset_first`: Soft reset the module before performing any other action. Default is `true`.
 
 Returns `true` or `false` depending on whether the activation was successful.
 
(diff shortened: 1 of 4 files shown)


johanstokking (commented):
In general it looks OK to me.

Regarding naming; we use lowerCamelCase, so it would be `resetFirst` and not `reset_first`.

@jpmeijers any comments?

jpmeijers:
I agree that the change looks OK, but I would want to test this change on a physical device. Unfortunately I do not have one at hand at the moment, so it will only be in a couple of days.

danalvarez:
>  
> Regarding naming; we use lowerCamelCase, so it would be `resetFirst` and not `reset_first`.
> 

@johanstokking I've updated the variable names to use camelCase.

@jpmeijers meanwhile, maybe you could assist with the "Help wanted" section of my original post in this PR?

jpmeijers (approved):
I've done the tests that I wanted, and they were successful.

The issue with only the prefix of the module number that is checked is not critical, and it can be fixed at a later stage in the `pgmstrcmp` function. The RN2xx3 modules are being deprecated, so I think we should anyway start limiting changes to this library.

